PR #23, opened 2011-03-22
Idna

xdecock:
Initial implementation to encourage Test and suggestions about the way to validate domains.

This implementation uses a large chunk of code from Maciej Lisiewski,

I'm waiting for GPL Confirmation, and way to add STD3-Ascii Validation (Whatever the f**k this is).

If you find informations about this, please post it.


gimler:
ping idn support for email addresses would be great

sstok:
From what I understand from the idn_to_ascii version notes when you use php 5.3 it should be supported already. https://secure.php.net/manual/en/function.idn-to-ascii.php

As new features are only accepted in the master branch?, it should be very easy to add support for this.

c960657:
Fixed in #1044.

PR #24, opened 2011-03-22
Amazon ses

xdecock:
Reimplementation of the basic class provided in bug #176

Exposes the whole API.

We require Amazon SDK to be loaded by the end User before starting the transport.

Exposes SendEmail and SendRawEmail.

I'll strongly recommend to store all Cloud API Email Providers into _Transports_Vendors_*

This will avoid Bloating the Transport_ Namespace :)

All suggestions are welcome.

### DIFF
--- a/docs/book/en/10-VendorTransport.markdown
+++ b/docs/book/en/10-VendorTransport.markdown
@@ -1,0 +1,48 @@
+Vendor Transports
+=================
+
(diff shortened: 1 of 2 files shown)


fabpot:
I'm closing this PR as Swiftmailer will be replaced by Symfony Mailer soon with support for third-party providers.

PR #187, opened 2012-04-19
Alternative file spool implementation.

williambailey:
Provides a new Swift_SerializedFileSpool class as an alternative to the standard Swift_FileSpool class.

Main benefits is that it can use the igbinary_serialize and gzcompress functions if available to further reduce the size of spool files. It also provides a bit more control over the generated spool files.

Most of the changes could have been incorporated into the existing spool class but there are are a few slight differences in implementation. Having a new class with slightly different naming conventions can also ensure that even if the two spool classes use he same directory they won't walk over each other.

### DIFF
--- a/lib/classes/Swift/SerializedFileSpool.php
+++ b/lib/classes/Swift/SerializedFileSpool.php
@@ -84,18 +84,18 @@ class Swift_SerializedFileSpool extends Swift_ConfigurableSpool
     public function queueMessage(Swift_Mime_Message $message)
     {
         $attempt   = 0;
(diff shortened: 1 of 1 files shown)


neufeind:
Could you please consider this patch? The TYPO3-team is "desperately" waiting for a fix ...
https://forge.typo3.org/issues/37594

fabpot:
I'm closing this PR as the Swiftmailer spool system does not work well and will soon be much better when we will replace Swiftmailer by Symfony Mailer.

Issue #185, opened 2012-04-19
UTF8 in header / suspicious characters in header / mails rejected


birdmedia:
Hello,

I’m using swiftmailer as part of Contao CMS where it is implemented for the mailing/newsletter system. 

I discovered that in cases where there are special characters (German Umlaute) used in the header, some recipients refuse to accept these mails because of “suspicious characters in header”.

Example header: Anfrage zu Ihren =?UTF-8?Q?Akti?=\015 =?UTF-8?Q?vit=C3=A4ten?=  („Aktivitäten“ would be correct)

Is this a kind of bug in the UTF8 encoding of swiftmailer? How to solve this problem?

Issue #79, opened 2011-09-02
subject utf-8 encoding in Outlook


swiftmailer:
I had a problem with subjects containing special characters. 
Characters are encoded correctly, but some extra linebreaks inserted to the subject header:

@@@ header
Subject: ugyeletes abcdef #37273-Closed-Low: =?utf-8?Q?=C3=81rv=C3=AD?=
 =?utf-8?Q?zt=C5=B1r=C5=91_t=C3=B6k=C3=B6rf=C3=BAr=C3=B3g=C3=A9p?=
@@@

These extra linebreaks doesn`t honored by some email clients. (Outlook Express 6.0.2900.5512)

I found a simple but ugly solution, by modifying the AbstractHeader.php: (included in symfony 1.4)

@@@ diff
--- Swift/Mime/Headers/AbstractHeader.php
+++ Swift/Mime/Headers/AbstractHeader.php
@@ -483,7 +483,7 @@
         '?' . $line . '?=';
     }
-    return implode("\r\n ", $encodedTextLines);
-    return implode(" ", $encodedTextLines);
  }
  
  /**
  @@@

Any idea for a cleaner solution?

Original creation date: 2010-03-01T13:11:15Z
Original reporter: Reilly Sweetland
Original ticket: http://swiftmailer.lighthouseapp.com/projects/21527/tickets/134

Issue #70, opened 2011-09-02
base64 encoding can wrong with utf8 characters on chunk boundry


swiftmailer:
It can happen that, in for example the subject of an email, an utf8 character falls "between" 2 chunks (the first byte is in the first chunk, the second byte is in the second chunk) of the base64 encoded string. All email clients that I have tested (thunderbird, nokia email) don't handle this. They will show an unknown character image instead.

A (real world) example (used in the subject of a mail):

Reminder: Invitation à participer à une enquête de satisfaction

What thunderbird shows is this: "Reminder: Invitation à participer �" (it cuts off the subject)
The encoded header looks like this:

Subject: =?utf-8?B?UmVtaW5kZXI6IEludml0YXRpb24gw6AgcGFydGljaXBlciDD?=
 =?utf-8?B?oCB1bmUgZW5xdcOqdGUgZGUgc2F0aXNmYWN0aW9u?=

I'm currently using swiftmailer 3.3.2, but by looking at the code I think 4.0.4 has the same problem.

Original creation date: 2009-09-04T10:31:57Z
Original reporter: E. de Ruiter
Original ticket: http://swiftmailer.lighthouseapp.com/projects/21527/tickets/108

PR #190, opened 2012-04-26
Fix email subject with special characters (e.g. UTF-8) in Outlook

aschempp:
Following up & fixing #70, #79 and #185. Outlook 2003 and later does not correctly handle "folding" in the subject, resulting in a totally broken emails (see http://cloud.iserv.ch/G9GF).

I removed the code folding entirely. Interesting is that "folding" is actually only applied to strings that need encoding, that is why the issue never appeared with ASCII-only subjects and only happens if your subject happens to match a certain string length where the encoded part is folded.

This might also be threaded as a SwiftMailer bug, but because it was not taken care of as of now, I think the best solution (to support Outlook) is not to apply it on encoding either.

### DIFF
--- a/lib/classes/Swift/Mime/Headers/AbstractHeader.php
+++ b/lib/classes/Swift/Mime/Headers/AbstractHeader.php
@@ -381,14 +381,9 @@ abstract class Swift_Mime_Headers_AbstractHeader implements Swift_Mime_Header
         )
       );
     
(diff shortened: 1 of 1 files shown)


aschempp:
Could this have been fixed in the latest version?
https://github.com/swiftmailer/swiftmailer/blob/v5.4.2/CHANGES#L15

fabpot:

Issue #222, opened 2012-07-03
default ssl/tls options ignored


rot13:
I always set verify_peer=TRUE and provide proper certificates when using SSL functions in PHP, but it doesn't work with SwiftMailer - default stream context options are ignored.

A simple fix could be:

```
diff --git a/lib/classes/Swift/Transport/StreamBuffer.php b/lib/classes/Swift/Transport/StreamBuffer.php
index d9801fa..ddccf9b 100644
--- a/lib/classes/Swift/Transport/StreamBuffer.php
+++ b/lib/classes/Swift/Transport/StreamBuffer.php
@@ -255,7 +255,7 @@ class Swift_Transport_StreamBuffer
     {
       $timeout = $this->_params['timeout'];
     }
-    $options = array();
+    $options = stream_context_get_options(stream_context_get_default());;
     if (!empty($this->_params['sourceIp']))
     {
        $options['socket']['bindto']=$this->_params['sourceIp'].':0';
```

Unfortunately it may break existing scripts relying on current behaviour, but at least it would be consistent with the rest of PHP, i.e. the programmer has to either supply a context or the default is used. I haven't found any way to supply my own context to SwiftMailer, so at least defaults should be respected. Otherwise it seems impossible to verify the certificate, which makes SSL/TLS support mostly useless (maybe even dangerous due to false sense of security).

PR #225, opened 2012-07-04
Respect default SSL/TLS options

adbre:
Pull request for issue #222

I'm not sure if this change should be opted in using a legacy setting or simlar.

### DIFF
--- a/lib/classes/Swift/Transport/StreamBuffer.php
+++ b/lib/classes/Swift/Transport/StreamBuffer.php
@@ -255,7 +255,7 @@ class Swift_Transport_StreamBuffer
     {
       $timeout = $this->_params['timeout'];
     }
-    $options = array();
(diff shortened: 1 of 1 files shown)


dhm80:
You opened up this pull request that lets us set SSL/TLS options. A great thing @adbre - especially because PHP does not seem to use these defaults if it not explicitly told to - but this might not always be flexible enough.

@acinelli made another approach to the same problem and added getter / setter methods. Also a great thing - but the default options you can set via stream_context_set_default() are ignored.

So how about a combination in form of "we merge the defaults and own settings"? What do you think?

dhm80:
I just reported the bug you are trying to patch at https://bugs.php.net/bug.php?id=69319
I would appreciate if you'd vote for the bug ;)

fabpot:

PR #302, opened 2013-03-12
Update AbstractSmtpTransport.php

AlexBorovnev:
When exists config with Wildcard names for sites, then SMTP protocol dont't work, because domain incorectly


fabpot:

PR #387, opened 2013-11-07
Fix for hotmail's RFC5322 bounces when reasonably complicated sender nam...

bjdelange:
...es are used.


fabpot:

PR #402, opened 2013-12-05
PHP Mail Transport to allow all recipients in BCC

rquetiez:
It is sometimes usefull to send emails with all recipients as BCC.

The ESMTP transport does allow this usage, while the Mail transport doesn't.

In commit https://github.com/swiftmailer/swiftmailer/commit/476586b810a98d06be477fd01c4309c74993e1b4, the fix was to get rid of a fatal error in two situations: no recipient in 'TO' or missing 'Subject'. That fix then consisted in allowing a missing Subject, but sending an exception in case there is no recipient in TO.
My proposal is to allow both usage in the end.

Regards,
Romain


fabpot:
Closing in favor of #772

PR #436, opened 2014-02-28
XOauth2 authentification for SMTP servers.

vladimir-tosev-etondigital:
Perhaps it may be useful to add OAuth2.0 authentication that is implemented on Gmail and MS Outlook servers (I don't know for any other).
Here is infos:
  http://msdn.microsoft.com/en-us/library/live/dn440163.aspx
  https://developers.google.com/gmail/xoauth2_protocol


fabpot:

PR #446, opened 2014-03-25
Optional explicit setting of message recipients

kevinsmith:
I need to be able to send email to a particular address with the message instead appearing to be addressed to a different email address, and it would appear [I'm not alone in this need](https://groups.google.com/forum/#!topic/swiftmailer/xfjjqEVU-zQ). Simply put, I need to be able to have RCPT TO differ from the To and Cc headers. (If you've ever had your email forwarded from an old address to a new one, you've seen this in action.)

It actually wasn't terribly difficult to add, but I'm at a loss on how to add tests for this. I can't get the test suite to run without errors on the master branch, though it looks like it's doing fine running on Travis. It also appears the library is undergoing a conversion to PHPUnit in swiftmailer/swiftmailer#418, which I applaud, but it leaves me wondering if I should try to write tests using the old system or PHPUnit. A little guidance here would be most appreciated.

Anything else I need to do to make it possible for you to merge this in, just let me know. Happy to help!


pmithrandir:
Hello,

Any news on this post ?
We have a similar need and are unable to do it with the swift mailer lib... or we didn't find how to...

Pierre

fabpot:
Closing in favor of #530

fabpot:

Issue #393, opened 2013-11-25
Break in long headers leads to a broken subject-line


neufeind:
We experienced problems with a broken subject-line. Judging from reports from mail-recipients (which we weren't yet able to reproduce) it seems to after such an "invalid" header some email-programs seem to parse the rest of the header as part of the email-body leading to various other "funny" bugs. Using Thunderbird etc. I did not experience any problems.
Investigating further into this brought me to a subject-line like:

Subject: Winterfest 2013 - Das Spiegelzelt kehrt nach Salzburg =?utf-8?Q?zorr?=^M
 =?utf-8?Q?rrr=C3=BCck?=

Notice the ^M which "vim" displays (and went unnoticed in other tools). I noticed that at this point the linux comandline "mail"-utility also shows a space-character in the subject-line, right in the middle of a word. This is the only header that contains a ^M and removing it this for example fixes the problem with "mail".

We experienced the problem with SwiftMailer 4.0.x (shipped with current TYPO3 4.5.30 LTS-release) but also tested it exists equally throughout the 4.x-releases including the latest 4.3.1. I couldn't yet test this on 5.x.

Fix: in classes/Swift/Mime/Headers/AbstractHeader.php in getTokenAsEncodedWord() change
return implode("\r\n ", $encodedTextLines);
to just
return implode("\n ", $encodedTextLines);

Since the \r\n is expected in various parts of the encoding before imho this "fix" late in the chain here looks like the best solution.

I expect this might also be related to bug-reports like #185.

clicsoft:
Hi there - problem still not solve (V 5.4.1 used in Typo3 7.6.2). 
After using Fix-proposal from @neufeind it works. Thanks!

fabpot:
Can anyone test with 5.4.2?

fvosberg:
Yes, this should be fixed with my bugfix from #655. Maybe I can verify that tomorrow!

neufeind:
For others using SwiftMailer as part of TYPO3: Recent versions of 7 LTS (7.6.x) include the fix from #655 and should be fine. The 6.2 LTS-tree doesn't contain the fix because of an older SwiftMailer. So solutions would be one of the following:
* upgrade to a recent TYPO3 7 LTS
* on 6.2.x patch the files yourself
* make sure to only use shorter subject-lines

PR #549, opened 2014-10-30
[BUGFIX] Break in long headers leads to a broken subject-line

neufeind:
Resolves: #393


MorrisJobke:
We noticed this bug too. The patch in here solves the issue for us.  It was quite hard to find the issue here. It would be super nice to get this in :)

@GrahamCampbell Can this be retested on a more current version and maybe the tests fixed to actually solve this issue.

MorrisJobke:
> @GrahamCampbell Can this be retested on a more current version and maybe the tests fixed to actually solve this issue?

I'm also open to help out here ;)

fabpot:

Issue #207, opened 2012-06-05
Remote URL attachment encoding


jadell:
Some attachments (speifically, PDFs) seem to be corrupted when using a remote file path.  The following snippet demonstrates the issue with a random PDF I found, but the same problem happens for some PDFs that my app generates:

```
$message = Swift_Message::newInstance();
  ->setFrom('foo@example.com')
  ->setTo('bar@example.com')
  ->setSubject('Test attachment')
  ->setBody('Some attachment you requested');

$message->attach(Swift_Attachment::fromPath('http://www.irs.gov/pub/irs-pdf/fw4.pdf, 'application/pdf')
  ->setFilename('w4.pdf'));

$message->send();
```

I tracked the issue down into `Swift_Mime_ContentEncoder_Base64ContentEncoder::encodeByteStream()`. In the `while` loop that reads the stream, if you cat together all the `$bytes` before encoding, and dump them out to a file, it is a valid PDF. If you cat together all the `$encodedTransformed` and dump them out to a file, then try to base64_decode it, the decoding returns `false` which indicates the encoding is failing somehow.

My current workaround is to use `Swift_Attachment::newInstance(file_get_contents("http://example.com/random.pdf"))` but I think that `fromPath` should be able to handle this.

PR #553, opened 2014-11-05
Fix base64 encoding with streams

apinstein:
Fixes [gh-207].


fabpot:
Thank you @apinstein.

apinstein:
No problem! Thanks for merging.

Issue #443, opened 2014-03-18
Swift_Signers_DKIMSigner: Setting 8bit encoder breaks DKIM signature


tomsommer:
Doing ->setEncoder(Swift_Encoding::get8BitEncoding()); breaks the DKIM signature calculation.

Issue #139, opened 2011-12-16
beforeSendPerformed event is called twice with memory spool


craue:
Using the memory spool added in d44b7fcc4c2abcc4e305672adb29833bab520fae, `beforeSendPerformed` of a `Swift_Events_SendListener` is called twice.

craue:
I'm also having this issue with file spool (using the current version 5.4.0). The event is triggered the 1st time when generating the mail and storing it in a file, then the 2nd time when it's actually sent. I'm using this event to append some footer text to each mail, which is unfortunately done twice.

@fabpot: It's a rather old issue. Do you even consider this to be a bug (which will be fixed) or do I need to make up a workaround?

craue:
The status is `RESULT_PENDING` in both invokations, so I don't see a way to distinguish between spooling and sending.

chris-testing-github:
I concur with @craue, in the beforeSendPerformed method the status is RESULT_PENDING in both instances so there is no way to differentiate. 
I'm using a custom plugin with beforeSendPerformed to emogrify an email before it's sent. Unfortunately because of this (bug?) the process is happening twice and characters are encoded twice leading to undesirable characters.

fabpot:
Anyone willing to submit a PR?

rmznpydn:
I have the same problem and do I have an existing solution?

samdjstevens:
I am also experiencing this issue - would an acceptable fix be changing the event target that is dispatched from `beforeSendPerformed` to say `beforeSpoolPerformed` (and similar for `sendPerformed`) or would this be a BC?

anyx:
Do anyone knows how to avoid this issue now?

appaydin:
> Do anyone knows how to avoid this issue now?

Look, maybe it helps https://github.com/appaydin/pd-mailer/blob/master/SwiftMailer/SendListener.php

kiwidood:
I came across this issue today, the easiest solution I found was to create an array with the key/or value of $message->getId() and checking against that before doing anything with the message. Not sure if it's fool proof, or if it would work in all use cases but maybe this will help someone.

PR #556, opened 2014-11-14
fix issue #443 #139 spool send event result

umanit:
Here is a PR for issues #443 and #139.

We've added a new result status in SendEvents with code 0x11 (RESULT_PENDING | RESULT_SPOOLED). It seems logical to us because this status lies between those two status.

The SpoolTransport now set RESULT_SPOOLED instead of RESULT_SUCCESS in order to be able to make the difference between a spooled mail and a sent mail.

Unit tests fails, but it wasn't due to anything that we've added. The master branch unit tests was already broken when we cloned the project.


sfdumi:
Any news on this?

fabpot:
Thank you @umanit.

Issue #558, opened 2014-11-18
Using S/Mime certificates that has intermediate certificates in the certificate chain


janflora:
We are using the S/Mime signer when we send out emails using swiftmailer. Our issue is that our S/Mime certificate has a long certificate chain with several intermediate certificates.

Right now we've solved the problem by extending the SmimeSigner, since the "openssl_pkcs7_sign" method already supports extra certificates.

If there is another way around this issue, we would really like to hear about it. If not, we would be happy to submit a pull request.

Best regards,
Jan Flora
Penneo

Issue #431, opened 2014-02-22
Joining additional certificates to PKCS#7 signature (aka extracerts feature)


obert01:
Hi,

It seems that Swiftmailer doesn't allow to join additional certificates while sending S/MIME signed messages. Even if it is a rather unused openssl feature, I faced a situation where I had to join the certificate of an intermediate CA to the messages which my PHP application was sending. The goal of this feature is to help the lail client in validating complex certificate chains. 

This option is called "extracerts" in the PHP manual page for the "openssl_pkcs7_sign" function : http://fr2.php.net/openssl_pkcs7_sign

Here is a small patch to SMimeSigner.php that adds an optional $extraCertificates parameter to the "setSignCertificate" method of the signer. Feel free to include this patch in Swiftmailer. 

```
--- SMimeSigner.php.original	2014-02-22 00:57:47.055609051 +0100
+++ SMimeSigner.php.new	2014-02-22 01:01:53.955605406 +0100
@@ -20,6 +20,7 @@
 class Swift_Signers_SMimeSigner implements Swift_Signers_BodySigner
 {
     protected $signCertificate;
+    protected $extraCertificates;
     protected $signPrivateKey;
     protected $encryptCert;
     protected $signThenEncrypt = true;
@@ -90,10 +91,11 @@
      * @param string       $certificate
      * @param string|array $privateKey  If the key needs an passphrase use array('file-location', 'passphrase') instead
      * @param integer      $signOptions Bitwise operator options for openssl_pkcs7_sign()
+     * @param string       $extraCertificates Extra certificates to join to the mail
      *
      * @return Swift_Signers_SMimeSigner
      */
-    public function setSignCertificate($certificate, $privateKey = null, $signOptions = PKCS7_DETACHED)
+    public function setSignCertificate($certificate, $privateKey = null, $signOptions = PKCS7_DETACHED, $extraCertificates = null)
     {
         $this->signCertificate = 'file://' . str_replace('\\', '/', realpath($certificate));
 
@@ -106,6 +108,9 @@
             }
         }
 
+	if (null !== $extraCertificates)
+	  $this->extraCertificates = str_replace('\\', '/', realpath($extraCertificates));
+
         $this->signOptions = $signOptions;
 
         return $this;
@@ -152,6 +157,14 @@
     /**
      * @return string
      */
+    public function getExtraCertificates()
+    {
+        return $this->extraCertificates;
+    }
+
+    /**
+     * @return string
+     */
[...]

PR #561, opened 2014-11-20
minor #431 #558 Using S/Mime certificates that has intermediate certs

janflora:
This PR fixes the problem described in #431 and #558.


sstok:
@fabpot ping

fabpot:
Thank you @janflora.

PR #568, opened 2014-11-30
fixed autoloading of unit test classes

fabpot:
It works by chance (as PhpUnit load all files/classes in alphabetical order and abstract classes are loaded first), but when trying to run a single unit test, it might break without this fix.


stof:
a better way would be to use autoload-dev in the composer.json

GrahamCampbell:
:+1: @stof

PR #569, opened 2014-11-30
Bumped version to 6.0

fabpot:
This PR contains my work towards 6.0. For now, it's more about some easy work to drop "old" stuff.

Next steps:

 - [ ] replace the Swiftmailer even dispatcher by the Symfony one.


stof:
@fabpot a new major version should be the time to get rid of the weird service location done when calling each parent constructor, which is totally ugly.

GrahamCampbell:
> Why php 5.5? Is there any way we could keep php 5.4 support in 6.0, and drop it later?

Actually, I retract that. :+1: for php 5.5!

GrahamCampbell:
Any news on this @fabpot? This looks really nice as a starting point.

fabpot:
@GrahamCampbell I think I will need some help for the next steps. I will add some steps in the description so that everyone can take one and contribute to this 6.0 version.

fabpot:
I've done even better than that: I've created a 5.x branch and master is now the upcoming 6.0 version. So, submitting PRs is going to be much simpler. I'm going to create an issue to reference the "next steps".

Issue #571, opened 2014-12-10
Allow setting ssl connection options, e.g. self-signed certificates


acinelli:
I'm working on a project using swiftmailer and needed to allow use of a self-signed certificate.  I've completed the simple work on this and will be making a pull request shortly.

acinelli:
Code implement here:
https://github.com/acinelli/swiftmailer/tree/topic_571

I've submitted a Pull Request, awaiting...

joostfaassen:
:+1:

h-wang:
:+1:

rdpascua:
:+1:

lollypopgr:
:+1:

shulcsm:
Accept please

hcoat:
:+1: 
This is very needed.

sstok:
@acinelli can you open a pull-request with these changes? Note that you must rebase on the 5.x branch to update your feature branch.

deminy:
:+1: Wish this (or an alternative approach) could be accepted. The fix is very important for using Swift Mailer on PHP 5.6. Thanks

aerkanc:
+1 for me! Please fix!

antic-eye:
:+1:

giovkanata:
:+1:

greabock:
```
Swift_DependencyContainer::getInstance()
            ->register('transport.buffer')
            ->asNewInstanceOf(InsecureStreamBuffer::class)
            ->addConstructorLookup('transport.replacementfactory');
```

where InsecureStreamBuffer extends Swift_Transport_StreamBuffer

Braunson:
:+1:

piotrooo:
:+1:  waiting for release

mareckii:
:+1: waiting for release

snwFlake:
:+1: also waiting for release :)

simoncpu:
It's November 25, 2016, this patch has not yet been merged. :(

PR #572, opened 2014-12-10
Implement Issue #571

tcinelli:
Allow estmp connection options to be set.  Can be used to allow ssl / tls to use self-signed certificates with the following "options" set:  array('ssl' => array('allow_self_signed' => true, 'verify_peer' => false)).

Reference:
http://stackoverflow.com/questions/26896265/php-swiftmailer-using-starttls-and-self-signed-certificates

### DIFF
--- a/lib/classes/Swift/Transport/EsmtpTransport.php
+++ b/lib/classes/Swift/Transport/EsmtpTransport.php
@@ -159,7 +159,31 @@ class Swift_Transport_EsmtpTransport extends Swift_Transport_AbstractSmtpTranspo
     {
         return $this->_params['tls'] ? 'tls' : $this->_params['protocol'];
     }
-
+	
+    /**
+     * Sets the connection options.
+     *
+     * @param array $options
+     *
+     * @return Swift_Transport_EsmtpTransport
+     */
+    public function setOptions($options)
+    {
+    	$this->_params['options'] = $options;
+    
+    	return $this;
+    }
+    
+    /**
+     * Returns the connection options.
+     *
+     * @return array
+     */
+    public function getOptions()
+    {
+    	return $this->_params['options'];
+    }
+    
     /**
      * Sets the source IP.
      *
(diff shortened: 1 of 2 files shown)


dhm80:
You opened up this pull request that lets us set SSL/TLS options. A great thing @acinelli - but the default options you can set via `stream_context_set_default()` are ignored.

@adbre made another approach to the same problem and simply added the SSL/TLS default options that can be set via `stream_context_set_default()`. Also a great thing (especially because PHP does not seem to use these defaults if it not explicitly told to) - but this might not always be flexible enough.

So how about a combination in form of "we merge the defaults and own settings"? What do you think?

lib/classes/Swift/Transport/EsmtpTransport.php
```diff
@@ -159,7 +159,31 @@ public function getEncryption()
     {
         return $this->_params['tls'] ? 'tls' : $this->_params['protocol'];
     }
-
+
```
GrahamCampbell: no trailing whitespace please

lib/classes/Swift/Transport/EsmtpTransport.php
```diff
@@ -159,7 +159,31 @@ public function getEncryption()
...
+     *
+     * @return Swift_Transport_EsmtpTransport
+     */
+    public function setOptions($options)
+    {
+    	$this->_params['options'] = $options;
+    
+    	return $this;
+    }
+
```
GrahamCampbell: no trailing whitespace please

lib/classes/Swift/Transport/EsmtpTransport.php
```diff
@@ -159,7 +159,31 @@ public function getEncryption()
...
+    /**
+     * Returns the connection options.
+     *
+     * @return array
+     */
+    public function getOptions()
+    {
+    	return $this->_params['options'];
+    }
+
```
GrahamCampbell: no trailing whitespace please

fabpot:

PR #578, opened 2015-01-04
Remove the unused private methods

SofHad:


fabpot:
Good catch, thanks @SofHad.

PR #580, opened 2015-01-07
fix phpdoc regarding setFrom() method

endelwar:
setFrom() method accepts both string and array as first paramenter


fabpot:
Good catch, thanks @endelwar.

Issue #551, opened 2014-11-04
getBody on attachments sends and empty attachment


rabas:
The issue seems to be that calling `getBody` on `Swift_Mime_SimpleMimeEntity` before actually sending the email makes the given attachment empty. 

Body of the attachment is read from the file, and it is instance `Swift_ByteStream_FileByteStream`. So when I call `getBody` it does return a string containing the body of an attachment. But the attachment is empty.

I can reproduce this by calling `getBody` twice:

```
var_dump($attachment->getBody());
// Dumps a string that contains attachment body

var_dump($attachment->getBody());
// Dumps an empty string 
```

As far as I can tell, `read` method of `Swift_ByteStream_FileByteStream` does not reset read handle correctly, because when I call `$os->setReadPointer(0)` before return of the `_readStream` method of the `Swift_Mime_SimpleMimeEntity` body is correctly returned each time.

Edit: Swiftmailer version is Swift-5.2.2-DEV and PHP version is PHP 5.6.0

PR #585, opened 2015-02-02
Reset read pointer

rabas:
Please reference https://github.com/swiftmailer/swiftmailer/issues/551 for description.


fabpot:
Thank you @rabas.

PR #591, opened 2015-03-23
Fixed branch alias

GrahamCampbell:


composer.json
```diff
@@ -25,7 +25,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "5.4-dev"
+            "dev-5.x": "5.4-dev"
```
stof: this does not make sense. The version for the 5.x branch is not ``dev-5.x``.
and there is no point aliasing it as it is already comparable

GrahamCampbell: Yeh, I know. Can we rename the 5.x branch to `5.4`?

GrahamCampbell: Because 5.x is wrong IMO.

stof: why ?

GrahamCampbell: @stof Because 5.x is not the same as 5.4.x. Also, all other branches in this repo, and in symfony/symfony follow the convention of aliasing the minor version too.

GrahamCampbell:
@fabpot Can you rename the branch to 5.4 please so it actually correctly represents 5.4.x-dev, like we have for the 4.x branches, and like we have for for symfony.

fabpot:
No, this branch will host 5.4, and eventually 5.5 if needed.

GrahamCampbell:
Then we can just make a 5.5 branch or rename this one when needed. Otherwise, how does composer know if this is 5.4.x-dev or 5.5.x-dev, because that could matter for people.

PR #597, opened 2015-03-30
QuickFix : Removed a notice

xdecock:
PHP Notice:  Undefined index: sourceIp in /swiftmailer/lib/classes/Swift/Transport/EsmtpTransport.php on line 184


fabpot:
Thank you @xdecock.

PR #606, opened 2015-05-01
Add $code as second parameter to (Transport/Io/Swift)Exception classes

timewasted:
SwiftException extends \Exception, which takes $code as an optional second parameter.  Add support for this second parameter to SwiftException as well as IoException and TransportException.  This allows for things such as getting the status code of an error when sending a message fails.

### DIFF
--- a/lib/classes/Swift/IoException.php
+++ b/lib/classes/Swift/IoException.php
@@ -19,7 +19,7 @@ class Swift_IoException extends Swift_SwiftException
      * Create a new IoException with $message.
      *
      * @param string $message
-     * @param int $code
+     * @param int    $code
      */
     public function __construct($message, $code = 0)
     {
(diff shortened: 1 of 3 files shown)


lib/classes/Swift/IoException.php
```diff
@@ -19,9 +19,10 @@ class Swift_IoException extends Swift_SwiftException
      * Create a new IoException with $message.
      *
      * @param string $message
+     * @param int    $code
      */
-    public function __construct($message)
+    public function __construct($message, $code = 0)
```
fabpot: Actually, what about adding the full signature of \Exception by adding the previous exception as well?

```php
__construct($message = null, $code = 0, Exception $previous = null)
```

timewasted:
I added the third parameter, but apparently forgot to mention that on this pull request.  Whoops!

fabpot:
Thank you @timewasted.

PR #609, opened 2015-05-05
Make LoggerPlugin keep track of the exception's code and previous exception

timewasted:
When LoggerPlugin catches a thrown exception, it modifies $message, and throws a new Swift_TransportException.  Previously, it only provided $message to the Swift_TransportException constructor.  Now, it will also pass the code and previous exception, so no information will be lost.

Also, it will append the code to the log message.


fabpot:
Thank you @timewasted.

Issue #337, opened 2013-07-08
smtp port 487 with startls


Lechus:
Hi,
I cant get to work switmailer with smtp server on port 487 with startlts.
`$transporter = Swift_SmtpTransport::newInstance(Config::get('mail.server'), 487, 'tls')
->setUsername(Config::get('mail.user'))
                            ->setPassword(Config::get('mail.password'));
             $mailer = Swift_Mailer::newInstance($transporter);`

<< 554 5.7.1 lechus@domain.pl: Recipient address rejected: Access
denied

!! Expected response code 250/251/252 but got code "554", with message
"554 5.7.1 lechus@domain.pl: Recipient address rejected: Access
denied

SysAdmin says that it needs to login first with existing account. Is this issue with STARTTLS and swiftmailer or do I need to change something?

PR #611, opened 2015-05-29
added getLastUsedTransport in Swift_Transport_LoadBalancedTransport

poison:
When using LoadBalancedTransport, I need to know (for tracking) which Transport it actually used. So I added this method to retrieve it.


poison:
Travis is failing (Build #337) because of a segfault in phpunit, which is unrelated to this commit. Can it be restarted?

fabpot:
Looks good to me. Can you add some tests?

fabpot:
Thank you @poison.

PR #619, opened 2015-06-24
Separate message/content ID generator class

c960657:
I would like greater control over the Message-ID headers without having to specify it manually on each message using setId().

Currently the ID depends on the global $_SERVER['SERVER_NAME'] variable.

This PR adds an ID generator class that is injected into all relevant instances instead of the EmailValidator. This seems more natural. The EmailValidator is now a dependency of the new IdGenerator instead.

The Message-ID is written to my mail server log. Injecting my own custom ID generator would allow me to generate per-message message IDs containing some information about the message itself, e.g. indicating the type of message, and thus make it easier to identify the various emails in the log files.

### DIFF
--- a/lib/classes/Swift/Mime/IdGenerator.php
+++ b/lib/classes/Swift/Mime/IdGenerator.php
@@ -1,0 +1,63 @@
+<?php
+
+/*
+ * This file is part of SwiftMailer.
+ * (c) 2004-2009 Chris Corbyn
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Egulias\EmailValidator\EmailValidator;
+
+/**
+ * Message ID generator.
+ */
+class Swift_Mime_IdGenerator implements Swift_IdGenerator
+{
+    /**
+     * @param EmailValidator $emailValidator
+     * @param string|null    $idRight
+     */
+    public function __construct(EmailValidator $emailValidator, $idRight = null)
+    {
+        if ($idRight) {
+            $this->idRight = $idRight;
+        } else {
+            $this->idRight = !empty($_SERVER['SERVER_NAME']) ? $_SERVER['SERVER_NAME'] : 'swift.generated';
+            if (!$emailValidator->isValid('dummy@'.$this->idRight)) {
+                $this->idRight = 'swift.generated';
+            }
+        }
+    }
+
+    /**
+     * Returns the right-hand side of the "@" used in all generated IDs.
+     *
+     * @return string
+     */
+    public function getIdRight()
+    {
+        return $this->idRight;
+    }
+
+    /**
+     * Sets the right-hand side of the "@" to use in all generated IDs.
+     *
+     * @param string $idRight
+     */
+    public function setIdRight($idRight)
+    {
+        $this->idRight = $idRight;
+    }
+
+    /**
+     * @return string
+     */
+    public function generateId()
+    {
+        $idLeft = md5(getmypid().'.'.time().'.'.uniqid(mt_rand(), true));
+
+        return $idLeft.'@'.$this->idRight;
+    }
+}
--- a/lib/classes/Swift/IdGenerator.php
+++ b/lib/classes/Swift/IdGenerator.php
@@ -1,0 +1,22 @@
+<?php
+
+/*
+ * This file is part of SwiftMailer.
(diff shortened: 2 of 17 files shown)


lib/classes/Swift/Mime/IdGenerator.php
```diff
@@ -0,0 +1,63 @@
...
+     * @param EmailValidator $emailValidator
+     * @param string|bool    $idRight
+     */
+    public function __construct(EmailValidator $emailValidator, $idRight = false)
+    {
+        if ($idRight) {
+            $this->idRight = $idRight;
+        } else {
+            $this->idRight = !empty($_SERVER['SERVER_NAME']) ? $_SERVER['SERVER_NAME'] : 'swift.generated';
+            if (!$emailValidator->isValid('dummy@'.$this->idRight)) {
```
fabpot: Instead of depending on the emailValidator here, what about copy/pasting the regex directly here?

stof: the email validator does not use a simple regex. the logic is complex (which is precisely why the external dependency is used, both here and in Symfony)

lib/classes/Swift/Mime/IdGenerator.php
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Message ID generator.
+ */
+class Swift_Mime_IdGenerator implements Swift_IdGenerator
+{
+    /**
+     * @param EmailValidator $emailValidator
+     * @param string|bool    $idRight
+     */
+    public function __construct(EmailValidator $emailValidator, $idRight = false)
```
stof: please use ``null``, not ``false``

lib/classes/Swift/Mime/IdGenerator.php
```diff
@@ -0,0 +1,63 @@
...
+    {
+        if ($idRight) {
+            $this->idRight = $idRight;
```
stof: shouldn't the custom idRight be validated too ?

c960657: In the existing code, the validation is only used to check if the contents of $_SERVER['SERVER_NAME'] contains something sane (AFAICT it always does, unless it has been filled with garbage on purpose).

The ID will be validated when the message is sent (in Swift_Mime_Headers_IdentificationHeader). It seems most in line with the behaviour elsewhere in Swift to throw errors on send, not on setting values or instantiating objects, so I think we should leave this. But I do not have strong opinions about this.

lib/classes/Swift/Mime/IdGenerator.php
c960657: Done.

fabpot:
Thank you @c960657.

PELock:
I don't like forced $_SERVER['SERVER_NAME'], for example I send mail list newsletter from my administrator panel from Yii2 backend environment eg. https://admin.domain.com and this URL (even if it's protected) gets revealed in the message headers, some helper function like setIdDomain() would be helpful.

c960657:
@PELock Yes, that is the main use-case for this new feature. Use `->setIdRight('example.com')` to set your preferred domain.

luxifer:
Any chances to have this PR tagged? I send my mails from the console and I don't want a workaround generate a custom Message-Id

acasademont:
Hi @c960657 what would be the best way to access the IdGenerator from a Symfony app? Seems like Swiftmailer uses it's own DI container. and the Swift_Message class auto-injects that dependency, so I can't modify the domain before the Id is actually generated. (I can regenerate the ID, but that seems a little bit hacky)

c960657:
Just use Swift Mailers own DI container:
```
Swift_DependencyContainer::getInstance()
    ->register('mime.idgenerator')
    ->asSharedInstanceOf('Swift_Mime_IdGenerator')
    ->withDependencies(['example.com'])
```

GeraldPK:
For anyone having the same problem i had and getting here via a search engine: As far as I can tell the above is not correct. (Anymore?) Throws a `Swift_DependencyException`:

`Cannot lookup dependency \"example.com\" since it is not registered.`

What worked for me:

```
Swift_DependencyContainer::getInstance()
   ->register('mime.idgenerator')
   ->asSharedInstanceOf('Swift_Mime_IdGenerator')
   ->addConstructorValue('example.com');
```

PR #621, opened 2015-06-25
More robust loading of autoload.php

c960657:
Depending on how you install PHPUnit, vender/autoload.php may already be loaded by PHPUnit itself.

If you include a file with require_once more than once, the second and consecutive calls will return true. This causes the script to die with a fatal error.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,6 +23,11 @@
     "autoload": {
         "files": ["lib/swift_required.php"]
     },
+    "autoload-dev": {
+        "psr-0": {
+            "Swift_": "tests/unit"
+        }
+    },
     "extra": {
         "branch-alias": {
             "dev-master": "5.4-dev"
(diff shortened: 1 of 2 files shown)


sstok:
You can also add the autoload entry as dev, not sure if it works in this case (PSR-0 with underscored class).

```json
    "autoload-dev": {
        "psr-0": {
            "Swift_": "tests/unit"
        }
    },
```

GrahamCampbell:
Yeh, autoload-dev should be used here.

stof:
In any case, using require_once when you care about the return value is always wrong, as you have no guarantee to get the original return value (you would get ``true`` if it was already loaded previously)

c960657:
@GrahamCampbell Done

fabpot:
Thank you @c960657.

Issue #620, opened 2015-06-25
Quoted-printable encoder exceeds 76-character limit


c960657:
**Sample code**

```
$factory = new Swift_CharacterReaderFactory_SimpleCharacterReaderFactory();
$charStream = new Swift_CharacterStream_NgCharacterStream($factory, 'iso-8859-1');
$encoder =  new Swift_Encoder_QpEncoder($charStream);
$input = str_repeat('a', 74) . ' ';
$output = $encoder->encodeString($input);
```

**Expected result**

```
string(81) "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa=
=20"
```

**Actual result (longer than 76-character limit)**

```
string(77) "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa=20"
```

c960657:
... This addition to the acceptance test also exhibits the problem (same bug exists in both QpEncoder and QpContentEncoder):

````
--- a/tests/acceptance/Swift/Mime/ContentEncoder/QpContentEncoderAcceptanceTest.php
+++ b/tests/acceptance/Swift/Mime/ContentEncoder/QpContentEncoderAcceptanceTest.php
@@ -46,6 +46,10 @@ class Swift_Mime_ContentEncoder_QpContentEncoderAcceptanceTest extends \PHPUnit_
                         $encoded .= $bytes;
                     }
 
+                    foreach (explode("\r\n", $encoded) as $i => $line) {
+                        $this->assertLessThanOrEqual(76, strlen($line));
+                    }
+
                     $this->assertEquals(
                         quoted_printable_decode($encoded), $text,
                         '%s: Encoded string should decode back to original string for sample '.
```

PR #622, opened 2015-06-25
Only add soft line break when necessary

c960657:
In quoted-printable encoding, the encoding lines should not exceed 76 characters. If there are longer lines in the source text, a soft line break should be added.

It is only necessary to add soft line breaks when a line in the source exceeds 76 characters. However, Swift_Encoder_QpEncoder and Swift_Mime_ContentEncoder_QpContentEncoder add soft line breaks every 76 character, ignoring any line breaks already present in the source.

This behaviour does not violate the spec, but it makes the encoded text harder to read with the human eye. Of course most people don't do this, but it is annoying when debugging, or when using a very primitive mail reader. One of the design goals of quoted-printable is to make the encoded text human-readable (unlike base64 encoding).

With this PR, the line length counter is reset every time a line break is encountered in the source text.

The change is implemented symmetrically for QpEncoder and QpContentEncoder. However, I did not add a change to QpContentEncoderAcceptanceTest similar to that made to QpEncoderAcceptanceTest,   because that revealed another bug, #620.

### DIFF
--- a/lib/classes/Swift/Encoder/QpEncoder.php
+++ b/lib/classes/Swift/Encoder/QpEncoder.php
@@ -198,14 +198,25 @@ class Swift_Encoder_QpEncoder implements Swift_Encoder
             }
 
             $enc = $this->encodeByteSequence($bytes, $size);
-            if ($currentLine && $lineLen + $size >= $thisLineLength) {
+
+            $i = strpos($enc, '=0D=0A');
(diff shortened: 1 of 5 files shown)


fabpot:
What about merging this one in 5.x instead of master?

c960657:
Yes, seems like a better choice. I have created a new PR for the 5.x branch: #627

PR #623, opened 2015-06-26
Implement Issue #571

Amunak:
PR as requested by @sstok. I'd love to see this in swiftmailer soon, as this prevents us from switching to the latest version of PHP if we don't want to fork swiftmailer.


sstok:
The tests are failing. https://travis-ci.org/swiftmailer/swiftmailer/jobs/68444689#L188

Amunak:
Damn, I guess I'll reopen it with my own fork then.

Amunak:
@sstok: I just noticed that the code standards test fails mostly on lines not touched by the PR.
And I wanted to ask what default options should we use? I feel like PHP did a very bad thing when they changed the default as it potentially breaks a ton of mailers. We could just leave it empty as you have it now, but shouldn't there be some sensible defaults that don't depend on PHP?

PR #624, opened 2015-06-26
Adds methods to set custom stream context options; implements #571

### LINKED ISSUE #571: Allow setting ssl connection options, e.g. self-signed certificates

Amunak:
Another try at #623 to fix #571.

This allows people to set (or override) any of the stream context options, including the SSL context, which in turn means we can use swiftmailer with SMTP over SSL/TLS with self-signed or otherwise invalid certificates.


fabpot:
Thank you @Amunak.

PR #625, opened 2015-06-27
Represent Date header as DateTimeImmutable object

c960657:
Representing a Date header with a DateTimeInterface object seems more obvious than a current Unix timestamp.

I suggest using DateTimeImmutable internally and automatically converting (mutable) DateTime instances supplied by the user. This prevents changing the value with breaking the cache inside Swift_Mime_Headers_AbstractHeader.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -7,6 +7,8 @@ Changelog
  * removed the Swift_Transport_MailInvoker interface and Swift_Transport_SimpleMailInvoker class
  * removed the Swift_SignedMessage class
  * removed newInstance() methods everywhere
+ * methods operating on Date header now use DateTimeImmutable object instead of Unix timestamp;
+   Swift_Mime_Headers_DateHeader::getTimestamp()/setTimestamp() renamed to getDateTime()/setDateTime()
  * bumped minimum version to PHP 5.5
 
 5.4.2 (2015-XX-XX)
--- a/lib/classes/Swift/Mime/HeaderSet.php
+++ b/lib/classes/Swift/Mime/HeaderSet.php
@@ -24,12 +24,12 @@ interface Swift_Mime_HeaderSet extends Swift_Mime_CharsetObserver
     public function addMailboxHeader($name, $addresses = null);
 
     /**
-     * Add a new Date header using $timestamp (UNIX time).
+     * Add a new Date header using $dateTime.
      *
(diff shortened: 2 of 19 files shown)


lib/classes/Swift/Mime/HeaderSet.php
```diff
@@ -24,12 +24,12 @@
...
-     * Add a new Date header using $timestamp (UNIX time).
+     * Add a new Date header using $dateTime.
      *
-     * @param string $name
-     * @param int    $timestamp
+     * @param string            $name
+     * @param DateTimeInterface $dateTime
      */
-    public function addDateHeader($name, $timestamp = null);
+    public function addDateHeader($name, $dateTime = null);
```
fabpot: You should probably add the type hint here as well.

lib/classes/Swift/Mime/SimpleHeaderSet.php
```diff
@@ -74,15 +74,15 @@ public function addMailboxHeader($name, $addresses = null)
...
-     * Add a new Date header using $timestamp (UNIX time).
+     * Add a new Date header using $dateTime.
      *
-     * @param string $name
-     * @param int    $timestamp
+     * @param string            $name
+     * @param DateTimeInterface $dateTime
      */
-    public function addDateHeader($name, $timestamp = null)
+    public function addDateHeader($name, $dateTime = null)
```
fabpot: type hint?

fabpot:
Looks good to me. Can you add a note in the CHANGELOG mentioning how it affects current code?

c960657:
Type hints and changelog entry added.

CHANGES
```diff
@@ -7,6 +7,7 @@ Changelog
  * removed the Swift_Transport_MailInvoker interface and Swift_Transport_SimpleMailInvoker class
  * removed the Swift_SignedMessage class
  * removed newInstance() methods everywhere
+ * methods operating on Date header now use DateTimeImmutable object instead of Unix timestamp
```
fabpot: There is more than that as the method names changed as well

c960657: I have added a mention of the renaming from getTimestamp()/setTimestamp() to getDateTime()/setDateTime().

fabpot:
Thank you @c960657.

Issue #520, opened 2014-09-04
MailTransport remote code execution


mappu:
I think the MailTransport is vulnerable to this issue.

In SwiftMailer, the parameter passed to `setFrom()` is `sprintf`'d into `-f%s`. If there is attacker-controlled `setFrom()` calls (e.g. hosted remailer or hosted SaaS product), then by setting the email address to a parameter containing `-X` the shell parameter injection causes **remote code execution**.

As far as i can see, calls to `setFrom()` are not validated for correct RFC email format. It's a little difficult to say for sure, though, via five layers of indirection and interfaces.

Prior disclosure of this issue:

- http://securitysucks.info/exploit-phps-mail-to-get-remote-code-execution/
- http://www.reddit.com/r/netsec/comments/2fc0zy/exploit_phps_mail_to_get_remote_code_execution_cc/

fabpot:
Reverted in #829. It was wrong as mail() in PHP already escapes this argument.

PR #626, opened 2015-06-27
Escape command-line args to Sendmail

c960657:
As reported in #520, we should escape command-line arguments. It is sufficient to escape the email address, not the entire $_extraParams string.


fabpot:
Thank you @c960657.

fabpot:
Reverted in #829. It was wrong as mail() in PHP already escapes this argument.

PR #629, opened 2015-07-01
new static to new self fix

htonus:


GrahamCampbell:
Why so many PRs?

stof:
@GrahamCampbell because he is doing them from the web UI

GrahamCampbell:
@stof What does this actually fix?

KorsaR-ZN:
@GrahamCampbell :+1:

stof:
the only idea I have is that this would fix compatibility with PHP 5.2. but Swiftmailer bumped the min requirement to 5.3.3, so I don't see any valid reason for this change.

KorsaR-ZN:
I think it doesn't need merge

GrahamCampbell:
:-1:

PR #630, opened 2015-07-01
new static to new self fix

htonus:


GrahamCampbell:
:-1:

htonus:
I'm sorry for didn't leave explanation. The idea was yes backward compatibility with 5.1.x-5.2.x. All the rest of the code written with "new self". Only these three classes prevent Swiftmailer to be used with older PHP versions.
Many projects were written on the frameworks designed for PHP 5.1.x - 5.2.x . Many ofthem were customized to meet some specific requirements. And it is very resourceful to migrate to PHP > 5.3.3
So why not to let such projects to upgrade Swiftmailer to latest version. Only this small change required.
Thank you.

Issue #603, opened 2015-04-29
Class `Swift_Plugins_Logger`: Add a timestamp with microtime to the returned output


arathorn2005:
This is a feature request.

It would be nice if the output generated by the Logger Plugin would contain a timestamp with microtime.
#### Example:

```
2015-04-29 09:37:10.001 ++ Starting Swift_SmtpTransport
2015-04-29 09:37:10.004 << 220 smtp.whatever.org SMTP Service ready at Wed, 29 Apr 2015 09:37:10 +0200 
2015-04-29 09:37:10.052 >> EHLO [127.0.0.1] 
2015-04-29 09:37:11.123 << 250-smtp.whatever.org Hello [127.0.0.1] ([1.2.3.4]), pleased to meet you 250-HELP 250-VRFY 250-EXPN 250-DSN 250-SIZE 250 PIPELINING 
2015-04-29 09:37:11.126 ++ Swift_SmtpTransport started
2015-04-29 09:37:11.139 >> MAIL FROM: <whoever@whatever.org> 
2015-04-29 09:37:11.222 << 250 whoever@whatever.org... Sender OK 
2015-04-29 09:37:11.333 >> RCPT TO: <someoneelse@whatever.org> 
2015-04-29 09:37:11.444 << 250 someoneelse@whatever.org... Recipient OK 
2015-04-29 09:37:11.555 >> DATA 
2015-04-29 09:37:11.666 << 354 Enter message, end with "." on a line by itself 
2015-04-29 09:37:11.777 >> . 
2015-04-29 09:37:11.888 << 250 Message accepted for delivery 
2015-04-29 09:37:11.999 ++ Stopping Swift_SmtpTransport
```

This would make it possible to use the log for detecting time-related issues (aka slow smtp server and things like that).

fabpot:
Can you submit a PR?

PR #639, opened 2015-07-17
added timestamp output to Logger Plugin

arathorn2005:
solves #603.

I haven't added any more details to it (p.e. allowing to configure the output format, allowing a switch if the timestamp should be added at all)... feel free to criticize!


fabpot:

Issue #645, opened 2015-08-03
QpContentEncoderProxy doesn't pass charsetChanged() to safeEncoder


koku:
Another bug we encountered after using Swift with the iso-8859-15 charset is that calls to QpContentEncoderProxy::charsetChanged() aren't passed to the safeEncoder.

We fixed that by adding the following line in the charsetChanged() method:
$this->safeEncoder->charsetChanged($charset);

Without this line the encoder being used assumed utf-8 is used.

PR #646, opened 2015-08-03
Pass charsetChanged() call to the safeEncoder

koku:
As mentioned in issue #645 we've created a fix to pass the charsetChanged call over to the safeEncoder.


sstok:
Is it possible to provide a test to make this doesn't get broken in the future?

koku:
I'll try to do that tomorrow (if I have time for it)

fabpot:
Thank you @koku.

PR #651, opened 2015-08-21
Changed type hint of $address of setReplyTo()

digilist:
As this method also accepts an array, I changed the parameter type to mixed (similar to setTo etc.).


fabpot:
Thank you @digilist.

PR #653, opened 2015-08-23
Added PHP 5.6, PHP 7 and HHVM to the build matrix.

hhamon:


GrahamCampbell:
What?

.travis.yml
```diff
@@ -1,12 +1,5 @@
 language: php
 
-php:
```
GrahamCampbell: this should not be deleted

hhamon: Why?

GrahamCampbell: why would you delete it?

.travis.yml
```diff
@@ -1,12 +1,5 @@
...
-  - 5.3
-  - 5.4
-  - 5.5
-  - 5.6
```
GrahamCampbell: php 5.6 is ALREADY in the build matrix

.travis.yml
```diff
@@ -1,12 +1,5 @@
...
-  - hhvm
```
GrahamCampbell: hhvm is ALREADY in the build matrix

GrahamCampbell:
`7.0` would be MUCH better than `nighlty` because nightly is actually php 7.1.

GrahamCampbell:
> nighlty because nightly is actually php 7.1.

Or rather, it's meant to be according the original RFC that said the master branch would become 7.1 immaculately following the RC1 release of 7.0.

.travis.yml
```diff
@@ -19,7 +12,14 @@ script:
...
-  fast_finish: true
+    include:
+        - php: 5.3
+        - php: 5.4
+        - php: 5.5
+        - php: 5.6
+        - php: nightly
+        - php: hhvm
+    allow_failures:
+        - php: nightly
```
fabpot: Can you use 7.0 here instead?

.travis.yml
hhamon: I used the same syntax taken from the symfony/symfony travis file.

.travis.yml
hhamon: Done!

.travis.yml
sstok: Travis allows to use both the build-matrix (advanced) and/or the simple `php` versions list.
But if you already use the matrix its cleaner to not use a list also (it's less confusing :+1:  ).

hhamon:
I managed to reproduce the segfaults on PHP5.3 and PHP5.6 on my machine. I'm trying to understand why it happens though.

hhamon:
Back to green! Disabling GC prevents segfaults when running PHPUnit.

tests/bootstrap.php
```diff
@@ -2,6 +2,9 @@
 
 require_once dirname(__DIR__).'/vendor/autoload.php';
 
+// Disable garbage collector to prevent segfaults
+gc_disable();
+
 set_include_path(get_include_path().PATH_SEPARATOR.dirname(__DIR__).'/lib');
 
 \Mockery::getConfiguration()->allowMockingNonExistentMethods(false);
```
GrahamCampbell: no leading slash needed

phpunit.xml.dist
```diff
@@ -32,6 +32,6 @@
     </testsuites>
 
     <listeners>
-        <listener class="\Mockery\Adapter\Phpunit\TestListener"></listener>
+        <listener class="\Mockery\Adapter\Phpunit\TestListener"/>
```
GrahamCampbell: no leading slash needed either

tests/bootstrap.php
```diff
@@ -2,6 +2,9 @@
 
 require_once dirname(__DIR__).'/vendor/autoload.php';
 
+// Disable garbage collector to prevent segfaults
+gc_disable();
+
 set_include_path(get_include_path().PATH_SEPARATOR.dirname(__DIR__).'/lib');
```
GrahamCampbell: seems silly to use the path separator, fine to just use `/`

GrahamCampbell: especially since we use that slash style in the same line...

hhamon: `PATH_SEPARATOR` is not `DIRECTORY_SEPARATOR`. `PATH_SEPARATOR` contains `:` on Linux and Mac OS.

fabpot:
Thank you @hhamon.

.travis.yml
stof: @hhamon the Symfony used an explicit matrix entirely just because of some experiments with a different order of jobs (running component jobs first), which was then reordered without going back to the simpler way to specify versions

stof:
> Back to green! Disabling GC prevents segfaults when running PHPUnit.

A bug should be opened on PHP though. A segfault in the garbage collector is a serious PHP bug

hhamon:
We should ask julien what he thinks about it.

PR #655, opened 2015-08-26
Fixing crlf for subject using mail

fvosberg:
The line endings get fixed for headers and body, but without the subject. That breaks subjects wich are too long. Especially when they contain quoted printables


fabpot:
Can you add a test that covers this to avoid any future regressions?

fvosberg:
Thanks for your fast reply. That unit test should prevent future regressions.

fabpot:
Thank you @fvosberg.

neufeind:
Can't test this right now, but I expect that also fixes #393 right?

PR #660, opened 2015-09-03
fixed test suite for PHP7

fabpot:
WIP


stof:
@fabpot the fact that you created the branch in the main repo rather than a personal fork means that Travis runs 2 builds (for the branch and for the PR). you might want to cancel one of them to let the queue reach recent builds faster

fabpot:
@stof Travis runs 2 builds anyway; the other one would have been on my fork.

stof:
@fabpot not if you disable Travis in your fork, which is possible as the PR is tested already and so the fork does not really need CI on its own (and it would not delay the PR builds as the quota of concurrent builds is per repo owner)

fabpot:
I've fixed everything that fails on my machine. The other ones do pass on my laptop.

PR #661, opened 2015-09-03
Switch to the docker-based infrastructure on Travis

stof:
This infrastructure boots faster than the legacy one, and the swiftmailer CI setup does not seem to need sudo


fabpot:
Thank you @stof.

PR #664, opened 2015-09-15
Fixed properties not declared explicitly

deguif:
There were also some wrong property names used (eg `showLen` vs `_showLen`) that could be causing some bugs.


fabpot:
Thank you @deguif.

PR #671, opened 2015-10-01
Fix - Consecutive read calls in Swift_ByteStream_FileByteStream should have the same behaviour

jakub-zalewski:
On first read method call with not successful fopen, return value from fopen is assigned to reader and exception is thrown.
For example: if exception is catched, on another read call on this object exception is not thrown because false is assigned to reader and this value is returned.

### DIFF
--- a/tests/bug/Swift/BugFileByteStreamConsecutiveReadCallsTest.php
+++ b/tests/bug/Swift/BugFileByteStreamConsecutiveReadCallsTest.php
@@ -1,0 +1,19 @@
+<?php
+
+
+class Swift_FileByteStreamConsecutiveReadCalls extends  \PHPUnit_Framework_TestCase
+{
+    /**
+     * @test
+     * @expectedException \Swift_IoException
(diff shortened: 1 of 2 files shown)


tests/bug/Swift/BugFileByteStreamConsecutiveReadCallsTest.php
```diff
@@ -0,0 +1,19 @@
...
+
+class Swift_FileByteStreamConsecutiveReadCalls extends  \PHPUnit_Framework_TestCase
+{
+    /**
+     * @test
+     * @expectedException \Swift_IoException
+     */
+    public function shouldThrowExceptionOnConsecutiveRead()
+    {
+        $fbs = new \Swift_ByteStream_FileByteStream('tralala');
```
fon60: Instead of 'tralala' it should be 'non existing file'

fabpot:
Thank you @jakub-zalewski.

Issue #331, opened 2013-06-21
Swift_TransportException: Expected response code 250 but got code "", with message ""


arnaugm:
Suddenly I started to have this error very frequently in my Symfony 2.1 applications. In the tests I run in jenkins it happens every 4-5 times I run the tests. The application running in the production server is giving this error too as it's reflected in the logs. And it is not a matter of sending tons of emails, just after sending 3 or 4 the error appears. Then you have to wait a bit until it works again.

As it's quite random it's impossible to know exactly at what point I started to get the error, but checking the commits, I don't see any change in the composer.lock file for long, so no changes in the bundles, and I haven't touched any package in the ubuntu server.

I updated "symfony/swiftmailer-bundle" respecting the composer.lock file of Symfony 2.1 which says "2.1.*" and the bundle was updated to the version 2.1.11. Swiftmailer itself didn't change and the installed version is 4.2.2.
The error persists.

I read something about being php's old version fault but I didn't get what was considered old. The server is using PHP 5.3.10-1ubuntu3.6.

As the error message is empty I really don't have any clue on what is going on.

hufeng903:
i use this in yii2 . get this error :Swift_TransportException

Expected response code 250 but got code "", with message "",who can help me？

vinoVINI1:
i got  error  while  trying to send  email   anyone  can  plz  help to  shot out this issues 
Expected response code 250 but got code "530", with message "530-5.5.1 Authentication Required. Learn more at 530 5.5.1 https://support.google.com/mail/?p=WantAuthError w72-v6sm80663446pfa.26 - gsmtp "

PR #673, opened 2015-10-02
Fix for #331. Transporter was stopped after sending. This resolves Sw…

mustafaileri:
I know this issue is closed but this problem still continues.
This PR  resolves : Swift_TransportException: Expected response code 250 but got code "", with message ""

### DIFF
--- a/lib/classes/Swift/Mailer.php
+++ b/lib/classes/Swift/Mailer.php
@@ -89,7 +89,7 @@ class Swift_Mailer
             }
         }
         $this->_transport->stop();
-        
+
         return $sent;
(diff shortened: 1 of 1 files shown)


xdecock:
This is a bad idea as it will stop the connection after each mail, so users using persistant connections or doing direct connections will be blacklisted after a few emails.

A simple example is if you send to orange or wanadoo or free in france after 60 mails you'll be put on throttle list and blacklisted a few mails later.

I'll recommend avoiding this patch as it will break the compatibility and implement bad behaviour for bulk sender.

PR #675, opened 2015-10-06
If egulias/email-validator bundle exists then use EmailValidator for …

ashutoshsilverline:
…swiftmailer to validate address in MailBoxHeader

### DIFF
--- a/lib/classes/Swift/Mime/Headers/MailboxHeader.php
+++ b/lib/classes/Swift/Mime/Headers/MailboxHeader.php
@@ -343,12 +343,17 @@ class Swift_Mime_Headers_MailboxHeader extends Swift_Mime_Headers_AbstractHeader
      */
     private function _assertValidAddress($address)
     {
-        if (!preg_match('/^'.$this->getGrammar()->getDefinition('addr-spec').'$/D',
-            $address)) {
-            throw new Swift_RfcComplianceException(
-                'Address in mailbox given ['.$address.
-                '] does not comply with RFC 2822, 3.6.2.'
+        if (class_exists('\Egulias\EmailValidator\EmailValidator')) {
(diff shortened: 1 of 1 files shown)


stof:
The 5.x version of Swiftmailer is still compatible with PHP 5.2, and this will trigger syntax errors on 5.2.

The 6.x dev version of Swiftmailer (i.e. the master branch) already uses this library.

Nek-:
Btw @stof this lib is not as good as is the current check of Swiftmailer. :/

But this PR is definitively wrong.

ashutoshsilverline:
Hi @stof, Actually swiftmailer is not supporting accent. so that's why used EmailValidator if exists.

Nek-:
`mélanie@free.fr` is not valid.

`melanie@méninge.fr` is valid.

Swiftmailer is ok with that.

ashutoshsilverline:
Address in mailbox given [ashutosh@gmailé.com] does not comply with RFC 2822, 3.6.2.
500 Internal Server Error - Swift_RfcComplianceException
2:39
in vendor/swiftmailer/swiftmailer/lib/classes/Swift/Mime/Headers/MailboxHeader.php at line 356  -

ashutoshsilverline:
Address in mailbox given [melanie@méninge.fr] does not comply with RFC 2822, 3.6.2.
500 Internal Server Error - Swift_RfcComplianceException 
in vendor/swiftmailer/swiftmailer/lib/classes/Swift/Mime/Headers/MailboxHeader.php at line 356 -

Nek-:
@ashutoshsilverline this is fixed in swiftmailer 6.x (thanks to usage of another validator).

But there is still a problem as the validator doesn't detect `mélanie@free.fr` as wrong.

Please close this issue that is not accurate, the problem is on another project now.

Issue #677, opened 2015-10-07
_id of Swift_Message is not renewed after the message is send with Swift_MailTransport


InfoTracer:
When the Swift_MailTransport is used as a transporter and the same Swift_Message object is used to send a message the _id is not renewed after sending the message.
When the Swift_SendmailTransport or the Swift_SmtpTransport is used the _id get renewed after sending.

used version: v5.4.1

I noticed this because Exchange marks the mails as duplicate and only sends me 1 instead of 10 mails. This was when i was testing the speed of the transporters on a local postfix setup.

Code example:

``` php
        try
        {
            $time_pre = microtime(true);
            //$transport = Swift_SendmailTransport::newInstance('/usr/sbin/sendmail -bs');
            //$transport = Swift_SmtpTransport::newInstance('localhost', 25);
            $transport = Swift_MailTransport::newInstance();


            // Create the Mailer using your created Transport
            $mailer = Swift_Mailer::newInstance($transport);

            // Create a message
            $message = Swift_Message::newInstance('Wonderful Subject')
                ->setFrom(array('test@test.nl' => 'test'))
                ->setTo( array( 'test@test.nl' ) )
                ->setBody('Here is the message itself');


            for ($i = 1; $i <= 10; $i++) {
                dump($message);
                $result = $mailer->send($message);
            }
            $time_post = microtime(true);
            $exec_time = $time_post - $time_pre;
            dump($exec_time);
        }
        catch ( \Exception $ex )
        {
            dump($ex);
        }
```

InfoTracer:
closed issues because mail transporter will be removed in version 6

PR #680, opened 2015-10-14
Placing the "generateId" method outside the else statement

InfoTracer:
Placing the "generateId" method outside the else statement so it will always be called when a message is sent. And not only when it failed to sent the message.

See issue #677 for more info


voku:
info comment (maybe only for my future myself): In the "SendmailTransport"-Class "generateId" is only called with parameter "-f" ?? | https://github.com/swiftmailer/swiftmailer/blob/5.x/lib/classes/Swift/Transport/SendmailTransport.php#L142

InfoTracer:
Closed PR because mail transporter will be removed in version 6

PR #683, opened 2015-10-19
fix removing a header in SimpleHeaderSet

jannschu:
I got an unexpected result from my code (attachments and S/MIME encryption+signing): There were multiple Content-Transfer-Encoding headers.

Multiple Content-Transfer-Encoding headers are added. When calling remove() one time, the has() function does return false, because it checked for index 0 by default which does not exist. The problem is PHP's unset() function, which does not update indices.

I'm far from an expert, so sorry If my assumption of this behavior being wrong is premature and indeed expected.


fabpot:
Closing in favor of #778

PR #688, opened 2015-10-28
use LF instead of CRLF for wider compatibility

### LINKED ISSUE #393: Break in long headers leads to a broken subject-line

MorrisJobke:
* fixes #393 
* fails with postfix on CentOS 6 otherwise

Fix is from #549 - Hopefully fix the tests that are failing in #549

### DIFF
--- a/lib/classes/Swift/Mime/Headers/AbstractHeader.php
+++ b/lib/classes/Swift/Mime/Headers/AbstractHeader.php
@@ -393,7 +393,7 @@ abstract class Swift_Mime_Headers_AbstractHeader implements Swift_Mime_Header
             }
         }
 
-        return implode("\r\n ", $encodedTextLines);
+        return implode("\n ", $encodedTextLines);
     }
 
     /**
--- a/tests/unit/Swift/Mime/Headers/UnstructuredHeaderTest.php
+++ b/tests/unit/Swift/Mime/Headers/UnstructuredHeaderTest.php
@@ -236,8 +236,8 @@ class Swift_Mime_Headers_UnstructuredHeaderTest extends \SwiftMailerTestCase
         $header->setValue($nonAsciiChar);
 
         $this->assertEquals(
-            'X-Test: =?'.$this->_charset.'?Q?line_one_here?='."\r\n".
-            ' =?'.$this->_charset.'?Q?line_two_here?='."\r\n",
+            'X-Test: =?'.$this->_charset.'?Q?line_one_here?='."\n".
+            ' =?'.$this->_charset.'?Q?line_two_here?='."\n",
             $header->toString()
             );
     }


MorrisJobke:
Or maybe `PHP_EOL` should be used? http://stackoverflow.com/a/6360550

> Postfix expects EOL to be LF when picking up mail from sendmail on unix and replaces that with CRLF when sending. When it gets CRLF it still replaces the LF and we get CRCRLF.

MorrisJobke:
This would explain the seen behavior. Maybe Swiftmail should use LF if it uses sendmail with postfix :see_no_evil:

GrahamCampbell:
> Or maybe PHP_EOL should be used?

That would lead to inconstant behaviour across platforms.

MorrisJobke:
@GrahamCampbell Do you have a good idea how to fix the broken behavior on postfix?

GrahamCampbell:
Hmmm, maybe just change the implode statement?

MorrisJobke:
> Hmmm, maybe just change the implode statement?

To what?

sstok:
Headers must follow the standards, which require the usage of CRLF!
https://tools.ietf.org/html/rfc5322#section-2.2

> Header fields are lines beginning with a field name, followed by a
   colon (":"), followed by a field body, and terminated by **CRLF**.

If Sendmail expects something else, the sendmail transporter should be fixed instead.

MorrisJobke:
It turns out that the is an issue with the sendmail version -> close

PR #700, opened 2015-12-10
Update RedirectingPlugin.php

gonzalovilaseca:
When extending the class method ``_isWhitelisted`` there is no access to ``$this->_recipient`` or ``$this->_whitelist`` so it might be good to make the local properties protected.


stof:
-1 for that. Just use the public getter and setters. There is no need to turn properties themselves into extension points for which BC must be taken into account

PR #712, opened 2016-01-06
Added support for iso-8859-15 encoding.

sanderdh:
Added support for ISO-8859-15 to allow for correct encoding of for example the Euro sign.


fabpot:
Thank you @sanderdh.

PR #713, opened 2016-01-06
Deleted unused private method

edsonmedina:


fabpot:
Thank you @edsonmedina.

PR #721, opened 2016-01-26
use uniqid instead of getRandomString

Arkounay:
When there are many mails to send (via spooling), using uniqid improved memory usage during file creation in my environnements


fabpot:

Issue #723, opened 2016-02-03
Failed to send mail - Invalid Sendmail options


sebastianfeldmann:
The `MailTransport` is sending extraParams even if no returnPath, sender or from address is defined.

In this case
```
sprintf($this->_extraParams, $reversePath)
```
will return `-f` and the mail will not be sent with the following error message
```
sendmail: option requires an argument -- 'f'
```

So the extraParam `-f` param should only be applied is a `reversePath` is set.

PR #726, opened 2016-02-04
bug #723 fixed php mail extra params on missing reversePath

sebastianfeldmann:
Fixed issue #723
Check reversePath and extraParams before adding them to php mail.


fabpot:
Thank you @sebastianfeldmann.

PR #728, opened 2016-02-22
convert encryption protocol value to lower case

serieznyi:
I spent the whole day could not understand why I am unable to send mail and I get the next error
`Connection could not be established with host smtp.yandex.ru [Unable to find the socket transport "SSL" - did you forget to enable it when you configured PHP? #1]`
I soon realized that the Protocol specify the encryption systems in the upper register. I was very pleasantly surprised that no one has foreseen such a possibility. So I decided to fix it

### DIFF
--- a/lib/classes/Swift/Transport/EsmtpTransport.php
+++ b/lib/classes/Swift/Transport/EsmtpTransport.php
@@ -140,6 +140,10 @@ class Swift_Transport_EsmtpTransport extends Swift_Transport_AbstractSmtpTranspo
      */
     public function setEncryption($encryption)
     {
+        if (is_string($encryption)) {
+            $encryption = strtolower($encryption);    
+        }
+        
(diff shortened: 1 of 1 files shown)


lib/classes/Swift/Transport/EsmtpTransport.php
```diff
@@ -140,6 +140,10 @@ public function getTimeout()
      */
     public function setEncryption($encryption)
     {
+        if (is_string($encryption)) {
```
fabpot: The condition is not needed, the only supported argument is a string.

fabpot:
Thank you @serieznyi.

lib/classes/Swift/Transport/EsmtpTransport.php
fabpot: fixed

Issue #445, opened 2014-03-22
getmypid() has been disabled for security reasons


adis-io:
https://github.com/swiftmailer/swiftmailer/blob/master/lib/classes/Swift/Mime/SimpleMimeEntity.php#L85

https://github.com/swiftmailer/swiftmailer/blob/master/lib/classes/Swift/Mime/SimpleMimeEntity.php#L420

https://github.com/swiftmailer/swiftmailer/blob/master/lib/classes/Swift/Mime/SimpleMimeEntity.php#L689

From php getmypid() function page

> **_Warning**_
> Process IDs are not unique, thus they are a weak entropy source. We recommend against relying on pids in security-dependent contexts.

amineameur:
hello guys first i'am not prof on this things :')
i realy tried to read this shit but i din'tget the joke 
very simple im using password reset to email a link to my users to change there password
so in local everything goes but when i set my app on a server when i send mail i got this error
getmypid() has been disabled for security reasons
tell me what to do without philosophy

PR #735, opened 2016-03-11
Removed getmypid function

burakcakirel:
It's removed in previous commit https://github.com/swiftmailer/swiftmailer/commit/84ff290871c3436c3703ce15ec5e344da0b59472
refs #445

But reverted back due to boundary length limit 
https://github.com/swiftmailer/swiftmailer/commit/84ff290871c3436c3703ce15ec5e344da0b59472#commitcomment-8746773

However, the issue which is mentioned in #445 is still there. So, I removed getmypid function and left the hash algorithm as md5

### DIFF
--- a/lib/classes/Swift/Mime/SimpleMimeEntity.php
+++ b/lib/classes/Swift/Mime/SimpleMimeEntity.php
@@ -80,7 +80,7 @@ class Swift_Mime_SimpleMimeEntity implements Swift_Mime_MimeEntity
      */
     public function __construct(Swift_Mime_HeaderSet $headers, Swift_Mime_ContentEncoder $encoder, Swift_KeyCache $cache, Swift_Mime_Grammar $grammar)
     {
-        $this->_cacheKey = md5(uniqid(getmypid().mt_rand(), true));
+        $this->_cacheKey = md5(uniqid(mt_rand(), true));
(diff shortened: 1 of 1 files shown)


sstok:
I think we can use random_bytes() with md5 (sha1?) to get the least chance of duplicates.
To also make this work for older version we can use https://github.com/paragonie/random_compat (PHP 5.2+)

fabpot:
Reading the comment from 84ff290871c3436c3703ce15ec5e344da0b59472, this is wrong as it will re-introduce a bug.

c960657:
AFAICT the comments for 84ff290 mentions two different bugs:

1. `md5()` was replaced with `sha1()`, causing MIME boundaries to exceed the allowed length. Hash collissions is hardly an issue here, so it should be safe to keep using MD5, despite it being unsafe for crypto purposes.
2. It is mentioned that `uniqid()` is only unique per process (I don't know the source of this).

MIME bondaries only have to be unique within the message, so `uniqid()` should be sufficient here.

Message IDs should be globally unique. This is hard to guarantee, but if we throw in openssl_random_pseudo_bytes() or random_bytes(), we can't do much better.
Here is one recommended way of generating message IDs: https://www.jwz.org/doc/mid.html#3

PR #738, opened 2016-03-20
Fix constructing of Swift_LoadBalancedTransport

xilnick:
Class LoadBalancedTransport for constructing uses call_user_func_array which calls Swift_Transport_LoadBalancedTransport::__construct but Swift_Transport_LoadBalancedTransport class does not have explicitly defined constructor which causes PHP Warning in php 7 .
https://github.com/swiftmailer/swiftmailer/blob/5.x/lib/classes/Swift/LoadBalancedTransport.php#L25

Adding Swift_Transport_LoadBalancedTransport::__construct fixes this issue.


GrahamCampbell:
Better to just fix the bad code than to add this pointless constructor.,

fabpot:
fixed in 13d4a53

PR #747, opened 2016-04-06
Can't use Swift_Transport_FailoverTransport because it no longer has a constructor.

BrunoDeBarros:
When using `Swift_FailoverTransport::newInstance($transports);`, it no longer works because the failover transport no longer has a __construct, and Swift_FailoverTransport [tries to call it](https://github.com/swiftmailer/swiftmailer/blob/5.x/lib/classes/Swift/FailoverTransport.php#L26).

This happened in the ["[Swift] various dead code cleanups, simplifications and phpdoc fixes"](https://github.com/swiftmailer/swiftmailer/commit/dae3bc6c530b86bcaca97ff2cf171b43328d569d) commit.


BrunoDeBarros:
Removing this until I have time to write a unit test to confirm all is working.

PR #749, opened 2016-04-20
Retry sending messages in the memory spool before aborting in case of transport failure

Seldaek:


lib/classes/Swift/MemorySpool.php
```diff
@@ -16,6 +16,7 @@
 class Swift_MemorySpool implements Swift_Spool
 {
     protected $messages = array();
+    protected $flushRetries = 3;
```
stof: This should be private instead (all new properties are added as private now, to avoid increasing the BC surface, even though lots of older properties still exist as protected for BC reasons)

fabpot:
Thank you @Seldaek.

PR #750, opened 2016-04-21
Add a return value description to Swift_Mailer::send method

vhraban:
I was trying to figure out what are the possible return types for send method, and send method PHPDoc was not helpful at all. I had to go through documentation and look for that, but it would be so much easier if the comment was a little bit more descriptive.

**Feature**: Add a description for Swift_Mailer::send method return type.

I hope this PR will help people like me and will save someone those precious couple of minutes


vhraban:
Travis says the tests have failed, but they have nothing to do with this change. Instead it seems like a travis build process is broken 
The command "gem install mailcatcher" failed and exited with 1 during .

fabpot:
Thank you @vhraban.

PR #751, opened 2016-04-29
Fix Travis Build

madflow:
The mailcatcher gem has a dependency that tries to install mime-types>=3 and needs Ruby >= 2.0. Travis bails out due to the Ruby Version. This PR should fix this.


fabpot:
Thank you @madflow.

PR #753, opened 2016-05-04
Prevent warning from being thrown if escapeshellarg is disabled within php.ini

steadweb:


c960657:
Why would one disable `escapeshellarg()`?

steadweb:
Locally we don't, but in production we disable various CLI functions. Seems like a common issue for people using shared hosting.

c960657:
If this function is commonly disabled for whatever reason, I agree it is worth considering to implement a polyfill. Any idea how widespread this configuration is?

I just don't understand why anyone would want to disable `escapeshellarg()` specifically. It is just a string function with no side effects.

steadweb:
No idea about widespread, but SO suggested others have faced this issue in regards to shared hosting. We maintain our production env, but getting this change through wouldn't be easy.

Our current fix is to rely on `5.4.1`.

c960657:
@steadweb If `escapeshellarg()` is disabled in your own production env, I suppose you know why this is the case? I would just like to understand the rationale behind this. It must be recommended somewhere, if it is common in shared hosting.

steadweb:
@c960657 I don't administer our servers, nor do I understand why this function has been disabled. Apologies I cannot answer that question for you. SO suggests this function isn't completely secure, but as you've pointed out (and I agree), this function simply escapes a string.

lib/classes/Swift/Transport/MailTransport.php
```diff
@@ -247,9 +247,32 @@ private function _getReversePath(Swift_Mime_Message $message)
...
+     * @return string
+     */
+    private function _escapeCommand($command)
+    {
+        if (is_string($command)) {
+            if (function_exists('escapeshellarg')) {
+                $command = escapeshellarg($command);
+
+            } else {
+                $command = str_replace('\'', '\\\'', $command);
```
stof: That's the wrong escaping logic (for instance on windows)

lib/classes/Swift/Transport/MailTransport.php
```diff
@@ -247,9 +247,32 @@ private function _getReversePath(Swift_Mime_Message $message)
...
+     * Escapes a string using either escapeshellarg or wrapping the command in single quotes if the function
+     * has been disabled.
+     *
+     * @param string $command
+     *
+     * @return string
+     */
+    private function _escapeCommand($command)
+    {
+        if (is_string($command)) {
```
stof: this check is useless as this method always expects a string as argument

fabpot:
I'm closing this PR as there is no more feedback and anyway, providing a secure/alternative implementation of such core PHP features as `escapeshellarg` seems not like a good idea.

steadweb:
@fabpot I agree it's not ideal to provide a polyfill for core PHP functionality, but this stops people upgrading passed 5.4.1. If I had the power / control to remove the function from the disabled functions setting, I would, because I agree, it's a string function and no more.

Another example of the polyfill being applied in a symfony component. 

https://github.com/symfony/process/pull/14/files

lib/classes/Swift/Transport/MailTransport.php
steadweb: Are you referencing the param type within the docblock? I was being defensive to make sure a string was passed.

lib/classes/Swift/Transport/MailTransport.php
steadweb: Could you suggest logic which would work across both platforms. It's also been PR'd here -https://github.com/symfony/process/pull/14/files

PR #755, opened 2016-05-07
Add Swift_Transport::ping()

c960657:
I have a long-running PHP job that occasionally sends email.

Unless I explicitly stop the SMTP connection after each use, the SMTP server may eventually close the connection if it has not been used for an extended period. This will make the next send() operation throw an exception.

This patch adds a ping() method similar to that in [Doctrine\DBAL\Connection::ping()](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Connection.php#L1560).

Apart from the addition of ping() to the Swift_Transport interface, this change is backwards compatible and can be merged into the current release branch.

### DIFF
--- a/lib/classes/Swift/Transport.php
+++ b/lib/classes/Swift/Transport.php
@@ -33,6 +33,29 @@ interface Swift_Transport
     public function stop();
 
     /**
+     * Check if this Transport mechanism is alive.
+     *
+     * If a Transport mechanism session is no longer functional, the method
+     * returns FALSE. It is the responsibility of the developer to handle this
+     * case and restart the Transport mechanism manually.
+     *
+     * @example
+     *
+     *   if (!$transport->ping()) {
+     *      $transport->stop();
+     *      $transport->start();
+     *   }
+     *
+     * The Transport mechanism will be started, if it is not already.
+     *
+     * It is undefined if the Transport mechanism attempts to restart as long as
+     * the return value reflects whether the mechanism is now functional.
+     *
+     * @return bool TRUE if the transport is alive
+     */
+    public function ping();
+
+    /**
      * Send the given Message.
      *
      * Recipient/sender data will be retrieved from the Message API.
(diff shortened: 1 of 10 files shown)


GrahamCampbell:
> this change is backwards compatible and can be merged into the current release branch.

This change is breaking because of the addition of a method to a public interface.

You have also not targeting the current release branch, which is fine, since this is breaking.

lib/classes/Swift/Transport.php
```diff
@@ -33,6 +33,29 @@ public function start();
...
+     *      $transport->stop();
+     *      $transport->start();
+     *   }
+     *
+     * The Transport mechanism will be started, if it is not already.
+     *
+     * It is undefined if the Transport mechanism attempts to restart as long as
+     * the return value reflects whether the mechanism is now functional.
+     *
+     * @param bool TRUE if the transport is alive
```
GrahamCampbell: did you not mean returns?

lib/classes/Swift/Transport/AbstractSmtpTransport.php
```diff
@@ -231,6 +231,31 @@ public function stop()
     }
 
     /**
+     * Check if this Transport mechanism is still alive.
```
GrahamCampbell: You need to either copy over the phpdoc, or use the correct inherit doc stuff otherwise the description from the other doc isn't applied here

GrahamCampbell: only phpdoc tags are inherited by default, descriptions are not

lib/classes/Swift/Transport.php
c960657: Yes - fixed.

lib/classes/Swift/Transport/AbstractSmtpTransport.php
c960657: Done.

c960657:
>This change is breaking because of the addition of a method to a public interface.

True, but if the interface change is omitted, the rest can be merged without breaking BC.

GrahamCampbell:
BC doesn't matter because you've sent this to 6.x anyway.

c960657:
What is the procedure for getting this feature (modulo the interface change) into 5.x? A separate PR after this one has been merged?

fabpot:
Thank you @c960657.

PR #756, opened 2016-05-07
Fix test failures due to uasort() warnings

c960657:
`uasort()` triggers a warning if the callback invokes a mocked method without an explicitly defined return value.

This patch adds explicit return values for all possible comparisons, making it independant on any specific implementation of `uasort()`.


fabpot:
This PR is just about the second commit, right?

c960657:
Yes, my bad. I have removed the irrelevant commit from the branch.

fabpot:
Thank you @c960657.

Issue #634, opened 2015-07-07
Switch from mcrypt to more secure implementation


Hikariii:
libmcrypt is a dead project, unmaintained for ~8 years, last version 2.5.8 was released in February 2007!
Various well maintained alternative libraries exist:
- openssl
- nss (Network Security Service)
- etc.

See also http://blog.remirepo.net/post/2015/07/07/About-libmcrypt-and-php-mcrypt

xdecock:
You can probably propose a pull request easily as it's seems to only affect

lib/classes/Swift/Transport/Esmtp/Auth/NTLMAuthenticator.php  and the
related test.

I don't use it so not sure i won't mess up everything.

On Tue, Jul 7, 2015 at 10:11 AM, Sebastiaan de Rooij <
notifications@github.com> wrote:

> libmcrypt is a dead project, unmaintained for ~8 years, last version 2.5.8
> was released in February 2007!
> Various well maintained alternative libraries exist:
>
>    - openssl
>    - nss (Network Security Service)
>    - etc.
>
> See also
> http://blog.remirepo.net/post/2015/07/07/About-libmcrypt-and-php-mcrypt
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/swiftmailer/swiftmailer/issues/634>.
>



-- 
Xavier De Cock
GPG Fingerprint: 93CA EE3F 9F57 5BE1 AE4A  794D 3C74 CA9E E7A5 0C1B
GPG Id: 0xE7A50C1B

PR #757, opened 2016-05-07
Use OpenSSL instead of mcrypt

c960657:
Fix for #634.

I haven't tested this with an actual NTLM server, but the unit test still passes. Inspired by https://github.com/robinpowered/php-ntlm/tree/master/src/Robin/Ntlm/Crypt/Des.


GrahamCampbell:
Should go to 6.x not 5.x IMO.

c960657:
@GrahamCampbell Could you elaborate on the reasoning behind this? Or perhaps direct me to some description that explains what should go into which branch in this project?

The NTLM code already requires the OpenSSL extension to be installed, so this patch does not add any new requirements.

fabpot:
Thank you @c960657.

PR #759, opened 2016-05-10
Fix tests failing on PHP7

c960657:
With these changes, the test suite runs without errors on PHP7.

The change in Swift_CharacterStream_NgCharacterStream is due to this BC break: https://bugs.php.net/bug.php?id=69931

### DIFF
--- a/lib/classes/Swift/CharacterStream/NgCharacterStream.php
+++ b/lib/classes/Swift/CharacterStream/NgCharacterStream.php
@@ -215,7 +215,7 @@ class Swift_CharacterStream_NgCharacterStream implements Swift_CharacterStream
      *
      * @param int $length
      *
-     * @return integer[]
+     * @return int[]
      */
     public function readBytes($length)
     {
--- a/tests/acceptance/Swift/Mime/ContentEncoder/QpContentEncoderAcceptanceTest.php
+++ b/tests/acceptance/Swift/Mime/ContentEncoder/QpContentEncoderAcceptanceTest.php
@@ -11,6 +11,11 @@ class Swift_Mime_ContentEncoder_QpContentEncoderAcceptanceTest extends \PHPUnit_
         $this->_factory = new Swift_CharacterReaderFactory_SimpleCharacterReaderFactory();
     }
 
+    public function tearDown()
+    {
+        Swift_Preferences::getInstance()->setQPDotEscape(false);
+    }
+
     public function testEncodingAndDecodingSamples()
     {
         $sampleFp = opendir($this->_samplesDir);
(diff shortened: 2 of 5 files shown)


tests/acceptance/Swift/Mime/ContentEncoder/QpContentEncoderAcceptanceTest.php
```diff
@@ -145,7 +150,6 @@ public function testDotStuffingEncodingAndDecodingSamplesFromDiConfiguredInstanc
         Swift_Preferences::getInstance()->setQPDotEscape(true);
         $this->testEncodingAndDecodingSamplesFromDiConfiguredInstance();
         // Disable DotStuffing to continue
```
fabpot: the comment above should probably be moved or removed.

tests/unit/Swift/Events/SimpleEventDispatcherTest.php
```diff
@@ -133,3 +133,10 @@ private function _createDispatcher(array $map)
         return new Swift_Events_SimpleEventDispatcher($map);
     }
 }
+
+class DummyListener implements Swift_Events_EventListener
+{
+    public function sendPerformed(Swift_Events_SendEvent $evt)
+    {
+    }
+}
```
fabpot: missing return here.

lib/classes/Swift/CharacterStream/NgCharacterStream.php
```diff
@@ -172,6 +172,10 @@ public function read($length)
                 $ret = substr($this->_datas,
                         $this->_currentPos * $this->_map,
                         $len);
+                // This check is necessary as of PHP 7.
+                if ($ret === '' && $this->_currentPos * $this->_map == strlen($this->_datas)) {
```
fabpot: Can we just do something like this:

```
if ($this->_currentPos * $this->_map == strlen($this->_datas)) {
    $ret = false;
} else {
    $ret = substr($this->_datas, $this->_currentPos * $this->_map, $len);
}
```

fabpot: We should probably also add a comment to explain why we need this piece of code.

fabpot: Or even better, just keep the `substr()` like today but make sure that the argument do not overflow.

c960657: Found it!  Turns out there was a bug (not specific to PHP7) in Swift_CharacterReader_GenericFixedWidthReader.

tests/unit/Swift/Events/SimpleEventDispatcherTest.php
c960657: Fixed

fabpot:
Good catch, thanks @c960657.

PR #761, opened 2016-05-12
Dependency-inject idRight into IdGenerator

c960657:
Instead of accessing `$_SERVER['SERVER_NAME']` directly from `Swift_Mime_IdGenerator`, it would be cleaner to inject `idRight` through the dependency injection container.

This patch also avoids the need to inject the email validator into the IdGenerator. The validator is only used to validate the config value. Elsewhere in Swiftmailer configuration values are not subject to such thorough validation.


fabpot:
Thank you @c960657.

PR #769, opened 2016-06-14
allow latest mockery 0.9.x

remicollet:
Trying to understand why composer.json states "mockery/mockery": "~0.9.1,<0.9.4"

With 0.9.5 test suite raise tons of 

      BadMethodCallException: Method Mockery_8_Swift_KeyCache::clearKey() does not exist on this mock object

Of course, the bootstrap.php doesn't allowMockingNonExistentMethods, so changing this seems enough.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         "php": ">=5.3.3"
     },
     "require-dev": {
-        "mockery/mockery": "~0.9.1,<0.9.4"
+        "mockery/mockery": "~0.9.1"
     },
     "autoload": {
         "files": ["lib/swift_required.php"]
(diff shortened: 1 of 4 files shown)


remicollet:
Related to 65863b665073255fb3654e00ebb3729fb4c8ef57 (no much explanation in this commit)

composer.json
```diff
@@ -18,7 +18,7 @@
         "php": ">=5.3.3"
     },
     "require-dev": {
-        "mockery/mockery": "~0.9.1,<0.9.4"
+        "mockery/mockery": "~0.9.1"
```
GrahamCampbell: Should this not be changed to `~0.9.5` if 0.9.4 is broken?

remicollet: 0.9.4 seems ok to

stof:
The question is why the testsuite tries to mock non existent methods ? this is likely to be a bug in the testsuite

fabpot:
It does not call non-existent methods, but methods that are not explicitly added via `shouldReceive()`.

fabpot:
Thank you @remicollet.

PR #771, opened 2016-06-20
dirname(__FILE__) to __DIR__

cheprasov:
PHP 5.3 has `__DIR__`.
`__DIR__` is equivalent to `dirname(__FILE__)`, but it works in 4-5 times faster.


fabpot:
Good catch, thanks @cheprasov.

PR #772, opened 2016-06-30
Allow mail transport send for messages without To header

adrianolek:
This resolves the issue from swiftmailer/swiftmailer#402 which hasn't been merged due to missing tests.


fabpot:
Thank you @adrianolek.

Issue #763, opened 2016-05-18
Address with unicode chars combined with a colon broke header


OliverFriedrich:
Somthing like 
`$message->setFrom(['test@test.com' => ':Test with unicode ÖÄÜ']);`
becomes 
`From: :Test with unicode =?utf-8?Q?=C3=96=C3=84=C3=9C?= <test@test.com>` 
and breaks the mail header, as there is the unquoted colon in the resulting header field. If the unicode chars ÖÄÜ are omitted the result is quoted and functional:
`From: ":Test with unicode" <test@test.com>`

Tested version: current download from github.

dmaicher:
Maybe related to an issue I posted a while ago :cry: 

https://github.com/swiftmailer/swiftmailer/issues/650

dmaicher:
I proposed a fix here https://github.com/swiftmailer/swiftmailer/pull/774

Issue #650, opened 2015-08-21
To header incorrectly encoded?


dmaicher:
We stumbled across a problem when sending emails with a certain name:

```php
$message->setTo([
    'test@example.com' => 'Hotel "zum heiteren Ösi"'
]);
```

This leads to the email just being send to "Hotel". So no domain at all.

We are using the ```mail``` transport. This is the ```$to``` value actually passed to ```Swift_Transport_SimpleMailInvoker::mail```

```Hotel "zum heiteren =?utf-8?Q?=C3=96si=22?= <test@example.com>``` 

This looks weird to me?

The problem seem to be the quotes AND the german character Ö. With just the quotes or just the Ö it works fine.

Any ideas?

**Comment by dmaicher** [comment]
I did another test with smtp transport:

```php
<?php

require_once 'vendor/autoload.php';

$mailer = new Swift_Mailer(Swift_SmtpTransport::newInstance());

$message = Swift_Message::newInstance('test', 'test')
    ->setTo('test@example.com', 'Hotel "zum heiteren Ösi"')
    ->setFrom('test@example.com')
;

$mailer->send($message);
```

This leads to this being sent via port 25:

```
MAIL FROM:<test@example.com>
250 2.1.0 Ok
RCPT TO:<test@example.com>
250 2.1.5 Ok
DATA
354 End data with <CR><LF>.<CR><LF>
Message-ID: <7d17d2be15fdd7d1abb17c08f83db1d8@swift.generated>
Date: Mon, 24 Aug 2015 20:49:27 +0200
Subject: test
From: test@example.com
To: Hotel "zum heiteren =?utf-8?Q?=C3=96si=22?= <test@example.com>
MIME-Version: 1.0
Content-Type: text/plain; charset=utf-8
Content-Transfer-Encoding: quoted-printable

test
.
250 2.0.0 Ok: queued as 5116829D5
QUIT
```

Then I did another test:

```php
$message = Swift_Message::newInstance('test', 'test')
    ->setTo('test@example.com', mb_encode_mimeheader('Hotel "zum heiteren Ösi"', 'utf-8', 'Q'))
    ->setFrom('test@example.com')
;
```

Which leads to this being sent:

```
MAIL FROM:<test@example.com>
250 2.1.0 Ok
RCPT TO:<test@example.com>
250 2.1.5 Ok
DATA
354 End data with <CR><LF>.<CR><LF>
[...]

Issue #206, opened 2012-06-05
Wrong encoding of address header


mschering:
If you have a from name like this:

Jonsson, Jørgen" jorgen.jonsson@example.com

Swift will encode it like this:

From: Jonsson, =?UTF-8?Q?J=C=B8rgen?= jorgen.jonsson@example.com

The mailserver will see this as two separate  mail addresses and it will turn it into:

From: Jonsson@mailserver.example.com, =?UTF-8?Q?J=C=B8rgen?= admin@intermesh.dev

This causes great confusion at my client!

I hope you can help us with a quick fix.
Thanks in advance,

Merijn Schering
Intermesh

PS: greater then and lower then signs are removed somehow

PR #774, opened 2016-07-06
Bugfix for broken MIME header encoding with quotes/colons and non-ascii chars

dmaicher:
This PR fixes bugs https://github.com/swiftmailer/swiftmailer/issues/763 and https://github.com/swiftmailer/swiftmailer/issues/650.

It is quite similar to the fix for bug https://github.com/swiftmailer/swiftmailer/issues/206 which has been merged with https://github.com/swiftmailer/swiftmailer/pull/211.

The problem again was that when a token contains valid alphanumeric characters and/or some RFC 2822 "specials" and there are other tokens with special characters then the whole header value was not quoted (as the special characters were already properly Q-encoded) and so the "specials" were left unescaped.

### DIFF
--- a/lib/classes/Swift/Mime/Headers/MailboxHeader.php
+++ b/lib/classes/Swift/Mime/Headers/MailboxHeader.php
@@ -277,9 +277,7 @@ class Swift_Mime_Headers_MailboxHeader extends Swift_Mime_Headers_AbstractHeader
      */
     protected function createDisplayNameString($displayName, $shorten = false)
     {
-        return $this->createPhrase($this, $displayName,
-            $this->getCharset(), $this->getEncoder(), $shorten
-            );
+        return $this->createPhrase($this, $displayName, $this->getCharset(), $this->getEncoder(), $shorten);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


lib/classes/Swift/Mime/Headers/AbstractHeader.php
```diff
@@ -234,15 +234,12 @@ protected function setFieldName($name)
...
-     * @param string                   $charset of the text
-     * @param Swift_Mime_HeaderEncoder $encoder
-     * @param bool                     $shorten the first line to make remove for header name
+     * @param string $string  as displayed
+     * @param bool   $shorten the first line to make remove for header name
      *
      * @return string
      */
-    protected function createPhrase(Swift_Mime_Header $header, $string, $charset, Swift_Mime_HeaderEncoder $encoder = null, $shorten = false)
+    protected function createPhrase($string, $shorten = false)
```
fabpot: You need to keep the signature as changing it would be a BC break. We can then clean that for 6.0.

dmaicher: True you are right :+1:  Changed it back.

dmaicher:
I also adapted the regex again because I found other issues with unescaped characters.

fabpot:
Thank you @dmaicher.

PR #779, opened 2016-07-09
updated to version 2.0 of the email validator dep

fabpot:


mrnnn:
Any idea on when you'll make this available in a release?

PR #781, opened 2016-07-15
Fix #426: Prevent fatal error during unserialization

### LINKED ISSUE #426: Unable to unserialize a broken Swift_Message
(This might just be a "feature" of PHP's serializing, but the fix I have found for it is within Swift Mailer so I'm reporting it here. Feel free to close if I'm off track!)

I am unable to unserialize a truncated serialized Swift_Message. Instead of `unserialize()` returning false and issuing a notice, a fatal error is coming from `SimpleMimeEntity::__destruct()`.

Example:

``` php
// Create and serialise a simple message.
$message = Swift_Message::newInstance()->setBody("body-start xxx bodyend");
$serialized = serialize($message);

// Chop the serialized message off halfway through the body.
$broken = substr($serialized, 0, strpos($serialized, 'xxx'));

// Try to retrieve the original object.
// An E_NOTICE should be issued, but instaead it's a fatal error.
$new = unserialize($broken);
```

The expected notice-error would be something like "Error at offset 6712 of 6727 bytes..." but instead it's "Call to a member function clearAll() on a non-object in \lib\classes\Swift\Mime\SimpleMimeEntity.php on line 834"

Is the following a possible fix for this?

``` php
public function __destruct()
{
    if (method_exists($this->_cache, 'clearAll')) {
        $this->_cache->clearAll($this->_cacheKey);
    }
}
```

Thanks for your time (and for producing a great library!).

PatchRanger:
## Description

Refer to #426 .
## Solution

I've looked for all of `__destruct()` implementations regarding cache usage. That was the only one - so it should completely fix the aforementioned issue.

### DIFF
--- a/lib/classes/Swift/Mime/SimpleMimeEntity.php
+++ b/lib/classes/Swift/Mime/SimpleMimeEntity.php
@@ -809,7 +809,9 @@ class Swift_Mime_SimpleMimeEntity implements Swift_Mime_MimeEntity
      */
     public function __destruct()
     {
-        $this->_cache->clearAll($this->_cacheKey);
+        if ($this->_cache instanceof Swift_KeyCache) {
+            $this->_cache->clearAll($this->_cacheKey);
+        }
     }
 
     /**


fabpot:
Thank you @PatchRanger.

PR #783, opened 2016-07-15
fixes #782: prevent fatal error in FileSpool.php when unserializing of message fails

### LINKED ISSUE #782: prevent fatal error in FileSpool.php when unserializing of message fails
FileSpool.php line 166-172:

`if (rename($file, $file.'.sending')) {
    $message = unserialize(file_get_contents($file.'.sending'));

```
$count += $transport->send($message, $failedRecipients);

unlink($file.'.sending');
```

  } else {`

When the unserialize fails $message becomes FALSE. If this is passed to the send() of a 
SendmailTransport.php which expects a Swift_Mime_Message, then a fatal error is thrown

> Catchable Fatal Error: Argument 1 passed to Swift_Transport_AbstractSmtpTransport::send() must be an instance of Swift_Mime_Message, boolean given [...]

This can be avoided by checking if the message is not FALSE before calling send (and deleting the '.sending'-file).

sneakyvv:
### Description

Refer to #782 .
### Solution

Check the result of unserialize before calling send(). Rename file to indicate unserialization error.

### DIFF
--- a/lib/classes/Swift/FileSpool.php
+++ b/lib/classes/Swift/FileSpool.php
@@ -164,15 +164,14 @@ class Swift_FileSpool extends Swift_ConfigurableSpool
 
             /* We try a rename, it's an atomic operation, and avoid locking the file */
             if (rename($file, $file.'.sending')) {
-            	$message = unserialize(file_get_contents($file.'.sending'));
-            	if ($message !== false) {
-            	    $count += $transport->send($message, $failedRecipients);
+                $message = unserialize(file_get_contents($file.'.sending'));
+                if ($message !== false) {
+                    $count += $transport->send($message, $failedRecipients);
 
                     unlink($file.'.sending');
-            	}
-            	else {
-            	    rename($file, $file.'.unserialize_error');
-            	}
+                } else {
+                    rename($file, $file.'.unserialize_error');
+                }
             } else {
                 /* This message has just been catched by another process */
                 continue;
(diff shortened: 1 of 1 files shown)


lib/classes/Swift/FileSpool.php
```diff
@@ -165,10 +165,13 @@ public function flushQueue(Swift_Transport $transport, &$failedRecipients = null
             /* We try a rename, it's an atomic operation, and avoid locking the file */
             if (rename($file, $file.'.sending')) {
                 $message = unserialize(file_get_contents($file.'.sending'));
+                if ($message !== false) {
```
fabpot: I would throw an exception instead.

chalasr:
Just faced this issue, @sneakyvv any chance that you finish this one? Otherwise I can take it over.

lib/classes/Swift/FileSpool.php
sneakyvv: @fabpot Wouldn't that cause the same problem, namely a fatal error? And thus causing all subsequent mails to not be sent. That seems not the desired behaviour.

sneakyvv:
@chalasr I've picked this up again. Sorry for the (very long) delay.

oytuntez:
We faced this same issue yesterday morning and our users received many duplicate emails. Any plans to fix this? Especially bothersome for bulk emails.

fabpot:

PR #785, opened 2016-07-25
Update MailTransport.php

cracki:
escapeshellarg() disabled on some shared hostings


GrahamCampbell:
I don't think it's a good idea not to escape if the function's disabled?

steadweb:
I provided a polyfill for this but it was deemed unnecessary. 

https://github.com/swiftmailer/swiftmailer/pull/753

I'll happily take suggestions to an alternative as we're unable to install the latest version of SwiftMailer at the moment.

steadweb:
Also https://github.com/swiftmailer/swiftmailer/pull/780

PR #789, opened 2016-08-19
Simple spelling fix

johncongdon:
BCMatch should be BCMath


fabpot:
Thank you @johncongdon.

Issue #794, opened 2016-08-30
Domain/Username "@" Issue in NTLMAuthenticator.php


nicksagona:
So, I've run into an issue if the username for the SMTP server doesn't have an "@domain.com" attached to it. The SMTP server I'm connecting to fails auth if you send it "username@domain.com" as the username and works if you use just "username".

However, if I do use just "username", the auth works and the mail goes through, but a notice gets thrown for "undefined offset" from this code, which tries to explode the string into an array based on the presence of "@":

https://github.com/swiftmailer/swiftmailer/blob/5.x/lib/classes/Swift/Transport/Esmtp/Auth/NTLMAuthenticator.php#L293-L302

Interestingly enough, if I change the "username" to just "username@" it actually works (auth works and mail is sent) and and there is no notice thrown.

So could a solution be to add a simple check there to see if there is a "@" present or not and then return the values accordingly (maybe $domain is null)?

PR #795, opened 2016-09-01
account for case of no domain passed with username in NTLMAuthenticator

shadiakiki1986:
Fixes [issue 794](https://github.com/swiftmailer/swiftmailer/issues/794)


fabpot:
Thank you @shadiakiki1986.

PR #798, opened 2016-09-11
Sanitize SERVER_NAME

c960657:
We need to sanitize `$_SERVER['SERVER_NAME']`.

The regexp is copied from Symfony\Component\HttpFoundation\Request.

This only affects master. The code in question was added in PR #760.


fabpot:
Thank you @c960657.

PR #799, opened 2016-09-11
Dependency-inject local-domain into AbstractSmtpTransport

c960657:
Instead of accessing `$_SERVER['SERVER_NAME']` directly from `Swift_Transport_AbstractSmtpTransport`, it would be cleaner to inject `localDomain` through the dependency injection container.

### DIFF
--- a/lib/classes/Swift/Transport/AbstractSmtpTransport.php
+++ b/lib/classes/Swift/Transport/AbstractSmtpTransport.php
@@ -38,12 +38,13 @@ abstract class Swift_Transport_AbstractSmtpTransport implements Swift_Transport
      *
      * @param Swift_Transport_IoBuffer     $buf
      * @param Swift_Events_EventDispatcher $dispatcher
+     * @param string                       $localDomain
      */
-    public function __construct(Swift_Transport_IoBuffer $buf, Swift_Events_EventDispatcher $dispatcher)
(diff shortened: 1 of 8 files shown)


lib/classes/Swift/Transport/AbstractSmtpTransport.php
```diff
@@ -458,34 +471,6 @@ private function sendBcc(Swift_Mime_Message $message, $reversePath, array $bcc,
...
-            $this->domain = $_SERVER['SERVER_NAME'];
-        } elseif (!empty($_SERVER['SERVER_ADDR'])) {
-            // Set the address literal tag (See RFC 5321, section: 4.1.3)
-            if (false === strpos($_SERVER['SERVER_ADDR'], ':')) {
-                $prefix = ''; // IPv4 addresses are not tagged.
-            } else {
-                $prefix = 'IPv6:'; // Adding prefix in case of IPv6.
-            }
-
-            $this->domain = sprintf('[%s%s]', $prefix, $_SERVER['SERVER_ADDR']);
```
fabpot: The logic of the `elseif` has been lost. Is it on purpose?

c960657: The handling of raw IP addresses has been moved to `setLocalDomain()`.

fabpot:
Thank you @c960657.

PR #803, opened 2016-09-25
Provide test cover and checks for null header files

ashtonherrington:
Currently there are no unit tests in place for cases where a header field is requested and has not been previously set.

In many places, arrays are expected to be returned, but the possibility of a null return value exists in cases where the header has not been set. No checks exist to confirm there is an array before attempting to save a value as its subscript, which will currently break.

Added checks if the header was not set to instead return an empty array.


fabpot:

PR #804, opened 2016-09-25
Adding the tests folder in .gitattributes

arcanedev-maroc:
This is not only for the distributed .zip archive but also to ignore the unwanted folders/files when we pull it with composer.

Usefull for clean/slim vendor folder (production env).


fabpot:
I'm closing this one as this was discussed multiple times in the past (mostly for Symfony) and rejected.

Issue #810, opened 2016-10-05
introduce a way to programatically disable the check _assertValidAddress in 


domnulnopcea:
Hi,

we have to introduce a way to disable the valid address check.

now what we do is to comment the code in the composer package. every update of swiftmailer will overwrite that 

gitlab.com is doing this also
https://gitlab.talview.com/resources/swiftmailer/commit/15b9ca896a2b4efb72cf0e7a9cead3a09536f3c7?view=parallel

thanks

voku:
Why do you need to comment out this check? Can't we fix this?

-> I also had problems with IDN domains and emoji in e-mail addresses, so I implemented a simple non RFC-valid check for this via https://github.com/voku/email-check/ with this commit https://github.com/voku/swiftmailer/commit/9be14951f638558a36c269b80da7bc9d187eae4d#diff-42584800e90c6d14774ed317b27bd069R346

domnulnopcea:
I need to comment it in order to send the email to addresses that do not pass the validation check

domnulnopcea:
any news here?

fabpot:
fixed in master

domnulnopcea:
Hi,
this https://github.com/swiftmailer/swiftmailer/pull/815/files is not fixing the issue.

it just removes a preg_match call with the Egulias\EmailValidator\EmailValidator implementation.

There should be provided a way to configure swiftmailer to skip this check all together.

big projects like gitlab are doing this but it prevents us from easily upgrading swiftmailer!

Issue #748, opened 2016-04-19
Address in mailbox given [tworzenieweb+hans.müller@gmail.com] does not comply with RFC 2822, 3.6.2.


tworzenieweb:
Hi,

i got the error like in subject for sending email to recipient contains special characters like ü.

``` php

    $message = Swift_Message::newInstance('some subject')
        ->setFrom('test@acme.com')
        ->setTo('tworzenieweb+hans.müller@gmail.com')
        ->setBody('test me');


    $result = $mailer->send($message);

```

Thanks

GrahamCampbell:
Why is this a bug? The email address is invalid.

xdecock:
not with utf-8 email address, but we welcome patches :)

On Tue, Apr 19, 2016 at 4:10 PM, Graham Campbell <notifications@github.com>
wrote:

> Why is this a bug? The email address is invalid.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/swiftmailer/swiftmailer/issues/748#issuecomment-211937200>
>



-- 
Xavier De Cock
GPG Fingerprint: 93CA EE3F 9F57 5BE1 AE4A  794D 3C74 CA9E E7A5 0C1B
GPG Id: 0xE7A50C1B

tworzenieweb:
Ok this is what I wanted to confront.

So is this format valid or no?

I read an RFC about it and it is pretty strict that this email address violates the rule but also there are some other validators like:

http://sphinx.mythic-beasts.com/~pdw/cgi-bin/emailvalidate

that said it is ok.

Also it is possible from gmail interface to send a valid email to this mailbox.

madflow:
@tworzenieweb As far as I know: https://tools.ietf.org/html/rfc2822 is obsoleted by https://tools.ietf.org/html/rfc5322, updated by https://tools.ietf.org/html/rfc5335, https://tools.ietf.org/html/rfc5336. Documented in #717 

@xdecock 

It is great to know that you do accept patches. At the moment this issue is the reason we consider maintaining a separate fork until this is resolved :S .

What is the favoured way to proceed? As far as I can see in the master branch there are some commits replacing the ```Grammar``` completely with a external script: https://github.com/egulias/EmailValidator.

* Would it be best to patch the current 5.4* release?
* Cherry-Pick/Integrate the commits from the master branch
* Wait for a 5.5 release?

Thanks

cerlestes:
[...]

Issue #717, opened 2016-01-22
Swift/Mime/Grammar.php is obsolete


djfd:
https://tools.ietf.org/html/rfc2822 is obsoleted by https://tools.ietf.org/html/rfc5322, updated by https://tools.ietf.org/html/rfc5335, https://tools.ietf.org/html/rfc5336

Particularly rfc5335, section-4.4 defines new grammar for addr-spec, allowing UTF8 for both local part and domain.

```
mailbox        =  name-addr / addr-spec / utf8-addr-spec

   angle-addr     =/ [CFWS] "<" utf8-addr-spec [ alt-address ] ">"
                     [CFWS] / obs-angle-addr

   utf8-addr-spec =  utf8-local-part "@" utf8-domain

   utf8-local-part=  utf8-dot-atom / utf8-quoted-string / obs-local-part

   utf8-domain    =  utf8-dot-atom / domain-literal / obs-domain

   alt-address    =  FWS "<" addr-spec ">"
```

This causes that current version cannot handle emails containing for example `ä` in local part (just my specific case) throwing an exception Swift_RfcComplianceException' with message 'Address in mailbox given does not comply with RFC 2822, 3.6.2.'

djfd:
there is my offer for improving this (tested against of my limited emails DB, does work, no any synthetic testes were done however, so can contain errors)
```
    public function init() {
 
        self::$_grammar['UTF8-tail'] = '[\x80-\xBF]';
        self::$_grammar['UTF8-2'] = '[\xC2-\xDF]'.self::$_grammar['UTF8-tail'];
        self::$_grammar['UTF8-3'] = sprintf('(?:\xE0[\xA0-\xBF]%s|[\xE1-\xEC]%s{2}|\xED[\x80-\x9F]%s|[\xEE-\xEF]%s{2})',
            self::$_grammar['UTF8-tail'],
            self::$_grammar['UTF8-tail'],
            self::$_grammar['UTF8-tail'],
            self::$_grammar['UTF8-tail']
        );
        self::$_grammar['UTF8-4'] = sprintf('(?:\xF0[\x90-\xBF]%s{2}|[\xF1-\xF3]%s{3}|\xF4[\x80-\x8F]%s{2})',
            self::$_grammar['UTF8-tail'],
            self::$_grammar['UTF8-tail'],
            self::$_grammar['UTF8-tail']
        );
[...]

Issue #419, opened 2014-01-17
Implement RFC 6531 (SMTP Extension for Internationalized Email)


InvisibleKind:
Is it possible to implement Proposed Standard RFC 6531?
Right now we're experincing a problems with German users, which has umplauts in their usernames. Like, " Address in mailbox given [äle@domain.org] does not comply with RFC 2822, 3.6.2.done "

We can't force them to change emails, because they are already using them and experience no problems at other services.

e1himself:
+1

MaxEndter:
+1

cerlestes:
:+1:

e1himself:
I've found a way to hack SwiftMailer to be able to work with i18ed email addresses.  
Check this gist: https://gist.github.com/e1himself/58c53aeef31a45345172  
We use this hack in our project, so it's verified it's working on SwiftMailer v5.2.2.

cerlestes:
What also seems to work, but requires PHP7, is to switch all PCRE-calls to UTF-8 mode and then use the new Unicode codepoint escape syntax to augment to `atext` block in SwiftMailer's Grammar class:

Grammar.php:84
```
self::$_grammar['atext'] = '[a-zA-Z0-9\u{0080}-\u{07ff}\u{0800}-\u{ffff}\u{10000}-\u{10ffff}!#\$%&\'\*\+\-\/=\?\^_`\{\}\|~]';
```

MailboxHeader.php:346
```
if (!preg_match('/^'.$this->getGrammar()->getDefinition('addr-spec').'$/Du',
```

This is basically what RFC 6531 is all about, but absolutely not the best way to do it. You'd have to rewrite SwiftMailer completely and make it more abstract for it to allow RFC 5321-based email messages (RFC 5321 = the current SMTP version from 2008; SwiftMailer is still based on RFC 2822 from 2001).

sstok:
No need for PHP7, the `Unicode codepoint escape syntax` is about strings not about regex.
The PCRE engine already supports unicode without any problems.

cerlestes:
@sstok: Yup, but those aren't included in `[a-z]`, obviously. The regular expressions used by SwiftMailer do not support Unicode (UTF-8) character sequences, so you have to augment them. One of the easier ways to do this, is to use unicode codepoints in PHP7 to specify the desired Unicode character ranges.

lenar:
[...]

PR #815, opened 2016-10-10
Use validator which supports RFC 6531

adeslade:
Addresses the following #419, #748, #810, #717 for 5.x. I know this has been addressed in master but I'm after PHP 5.3 support. 

I had an alternative approach in https://github.com/adeslade/swiftmailer/commit/01cb0870d3514e83327f6a8936cc97cbe03abb4f. I wasn't happy with this though as the `Swift_Mime_Headers_MailboxHeader` if manually instantiated would not have been backwards compatible.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,8 @@
         }
     ],
     "require": {
-        "php": ">=5.3.3"
+        "php": ">=5.3.3",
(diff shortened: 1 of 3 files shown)


fabpot:
This indeed fixed in master, but cannot be part of the 5.x series.

PR #825, opened 2016-11-13
Inconsequent clone method

stodorovic:
removing brackets for clone method in Swift_Message::__clone()

At all other places are clone without brackets according with PHP standards

Example:
https://github.com/swiftmailer/swiftmailer/blob/master/lib/classes/Swift/Mime/SimpleMimeEntity.php#L804

Related to https://github.com/wimg/PHPCompatibility/issues/284


fabpot:
Thank you @stodorovic.

PR #832, opened 2016-11-25
Fix phpdoc return tag in logger pluggin

vdechenaux:
Hi,

just a small fix for phpdoc


fabpot:
Good catch, thanks @vdechenaux.

PR #838, opened 2016-12-07
Remove deprecated code

### LINKED ISSUE #238: problems with mail() transport and safe_mode
(I know, `safe_mode` is deprecated and horrible anyway, but with some hosted webspaces, there is nothing you can do about it...)

In `lib/classes/Swift/Transport/SimpleMailInvoker.php` the mail() method looks like this:

``` php
    public function mail($to, $subject, $body, $headers = null, $extraParams = null)
    {
        if (!ini_get('safe_mode')) {
            return @mail($to, $subject, $body, $headers, $extraParams);
        } else {
            return @mail($to, $subject, $body, $headers);
        }
    }
```

I ran into a problem there: `ini_get()`

> Note: When querying boolean values
> 
> A boolean ini value of off will be returned as an empty string or "0" while a boolean ini value of on will be returned as "1". The function can also return the literal string of INI value. 

So in case this returns a string, e.g. "Off", this makes the condition behave wrongly, and fail (silently), since `safe_mode` does not allow for the additional parameter.

peter-gribanov:
PR for fix #238

### DIFF
--- a/lib/classes/Swift/Transport/SimpleMailInvoker.php
+++ b/lib/classes/Swift/Transport/SimpleMailInvoker.php
@@ -30,10 +30,6 @@ class Swift_Transport_SimpleMailInvoker implements Swift_Transport_MailInvoker
      */
     public function mail($to, $subject, $body, $headers = null, $extraParams = null)
     {
-        if (!ini_get('safe_mode')) {
-            return @mail($to, $subject, $body, $headers, $extraParams);
-        }
-
         return @mail($to, $subject, $body, $headers);
     }
 }


peter-gribanov:
This bug is reproduced on

```
Linux 3.10.0-327.36.3.el7.x86_64 x86_64
PHP 5.6.28
```

and

```
Windows 7 x64
PHP 5.6.24
```

fabpot:
Closing as 5.x still supports PHP 5.3. This has been removed in master though.

PR #839, opened 2016-12-10
Update .gitattributes

Aliance:


fabpot:
The changes you propose here and on some of my other repositories have been discussed more than once in the past and we decided to not do it. Sorry but I'm going to close them now. Thanks.

PR #842, opened 2016-12-20
Interfaces removal

fabpot:
This is an intermediary step to ease the next refactoring steps. Right now, it's difficult to reason about the current system as there are just too many (useless) interfaces/classes.


prisis:
Is it okay if i make a pr for psr-4?

fabpot:
@prisis That won't work with the current architecture.

prisis:
@fabpot Was hoping that the architecture get a refactoring too :)

PR #845, opened 2016-12-27
Added a renderAndSend() method to define email messages as Twig templates

javiereguiluz:
### Context

A while ago we proposed an alternative way to send emails in Symfony apps. See https://github.com/symfony/symfony/issues/15060  At that time, @stof cleverly spotted several problems related to Twig ... but the Twig 1.28 version introduced some features that probably will solve all of our problems.

### Proposal

I propose to add a `renderAndSend()` method as an alternative to `send()`. This pull request is a **super early proof of concept draft**. This is just to see if this idea is doable or stupid.

### Examples

For simple emails where everything is static, you can do this:

```php
$this->get('mailer')->renderAndSend('emails/registration.twig');
```

And the template:

```twig
{# app/Resources/views/emails/registration.twig #}
{% block to 'user@example.com' %}
{% block from 'donotreply@example.com' %}
{% block subject %}Welcome to ACME{% endblock %}

{% block body %}
    ...
    <div> ... </div>
    ...
{% endblock %}
```

If the template needs parameters, pass them as an array in the optional second argument:

```php
$this->get('mailer')->renderAndSend('emails/registration.twig', array(
    'name' => $name
));
```

This second argument also allows to define the value of some email message paramenters: `to`, `from`, `cc`, `bcc`, `subject`, `body` and `body_txt`:

```php
$this->get('mailer')->renderAndSend('emails/registration.twig', array(
    'from' => 'send@example.com',
    'to' => 'recipient@example.com',
    'subject' => 'Hello Email',
    'name' => $name,
));
```

### DIFF
--- a/lib/classes/Swift/Mailer.php
+++ b/lib/classes/Swift/Mailer.php
@@ -17,15 +17,17 @@ class Swift_Mailer
 {
     /** The Transport used to send messages */
     private $transport;
+    private $twig;
 
     /**
      * Create a new Mailer using $transport for delivery.
      *
      * @param Swift_Transport $transport
      */
-    public function __construct(Swift_Transport $transport)
+    public function __construct(Swift_Transport $transport, Twig_Environment $twig = null)
     {
         $this->transport = $transport;
+        $this->twig = $twig;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


sstok:
Should this not be done for the SwiftBundle? 😊

javiereguiluz:
@sstok honestly I don't know. This is what Fabien said about this in the related issue:

> That should be either part of Swiftmailer directory or the Swiftmailer bundle, but that cannot be in Symfony itself. I think I'd prefer something in Swiftmailer directly as it would be useful outside the context of Symfony as well then.

rvanlaak:
Although it is great to have Twig abilities in rendering `from`, `to` and `subject`, it feels a bit weird that these variables can be added by passing them to the template together with the data array. Especially as `Swift_Message` has getters/setters for that. So in other words, the DX on that could be improved as the interface now is kinda unclear.

javiereguiluz:
@rvanlaak the reasoning was that if you prefer the low-level creation of `Swift_Message` objects, you can use the `send()` method. This new method is for developers who don't care about the internal details: just get the `mailer` service, render a template and send the email. You even aren't aware of the existence of SwiftMailer, and that's OK (I guess).

rvanlaak:
@javiereguiluz well, your code sample at the bottom of the PR actually adds `from`, `to` and `subject` to one big data array for the rendering, the Django method signature explicitly lists them. That is my small concern, that it might look weird that `from`, `to` and `subject` aren't all parameters themselves but are part of the "big" data array.

In other words, looking at the Django example I'd as developer expect it to work like this:

```php
$this->get('mailer')->renderAndSend('emails/registration.twig',
    'send@example.com',
    'recipient@example.com',
    'Hello Email',
    ['name' => $name],
);
```
The `renderAndSend` in my opinion could merge the `from`, `to` and `subject` values in the `data` array so they actually are available to the Twig template though.

> You even aren't aware of the existence of SwiftMailer, and that's OK (I guess).
From Symfony point of view this would be awesome, but the proposal is to put this in `SwiftMailer` right? 👍

javiereguiluz:
I'm closing because this proposal stalled and it even has more downvotes than upvotes.

Issue #844, opened 2016-12-26
Security Issue


pizzeys:
There's a bit of noise going around at the moment regarding a vulnerability in PHPMailer where if an attacker can control the from address a script uses they can pass arbitrary arguments to the sendmail binary, which, if it's actually sendmail, results in being able to write to an arbitrary file (and so on most PHP setups, likely RCE) and if not at least allows to pass other parameters the developer perhaps didn't intend.

Swiftmailer is actually vulnerable to the same bug, if using the Swift_MailTransport transport - for example, one can do:

setFrom(array('"x@y.com\" -X/tmp/test"@test.php' => 'John Doe'))

Is this expected? It would be quite easy for a script to be using this in a dangerous way, so it should probably be made clear in the documentation that this parameter should be handled with kid gloves, if so.

fabpot:
Where you able to reproduce this with Swiftmailer? I didn't, so if you did, can you tell me how (email is probably better)? Thanks.

joserprieto:
The complete vulnerability is described here:

http://legalhackers.com/advisories/PHPMailer-Exploit-Remote-Code-Exec-CVE-2016-10033-Vuln.html

And the key concept is:

> Further research however shown that the validation is done using the RFC 3696 specification.
> 
> The RFC allows emails to contain spaces when quoted with ". 
> According to the spec, the following email address is valid:
> 
> " some test email address with spaces"@weird-email.com
> 

I can't test this right now, but, if you can send an email like this:

```
$email_from = '"attacker\" -oQ/tmp/ -X/var/www/cache/phpcode.php  some"@email.com';
$msg_body  = "<?php phpinfo(); ?>";
```

You can test the vulnerability.

Regards.

pizzeys:
It looks like this may have previously been fixed here:

https://github.com/swiftmailer/swiftmailer/issues/520

But then reverted due to 'mail() in PHP already escapes this argument.' However that's not the case - PHP passes the parameters through escapeshellcmd() but that doesn't prevent the appending of additional arguments, which is the issue here.

However, that may mean that the fix used there is appropriate for re-use here perhaps?

MartinZubek:
So how does it look like guys? Should we be worried?

z38:
Note that the first fix for CVE-2016-10033 by PHPMailer was incomplete. It seems one has to be quite careful when escaping the additional parameters (CVE-2016-10045):

[...]

PR #846, opened 2016-12-29
[Security][WIP] Add isSafeShell check before passing extra params to mail()

barryvdh:
Potential fix for #844

@Zenexer Are you okay with using the (slightly reformatted) code from https://github.com/PHPMailer/PHPMailer/pull/929 here? Do you agree this fixes the issue?


Zenexer:
@barryvdh Checking now.

lib/classes/Swift/Transport/MailTransport.php
```diff
@@ -239,6 +239,35 @@ private function _getReversePath(Swift_Mime_Message $message)
     }
 
     /**
+     * Fix CVE-2016-10074 by disallowing potentially unsafe shell characters.
+     *
+     * Note that escapeshellarg and escapeshellcmd are inadequate for our purposes, especially on Windows.
+     * @param string $string The string to be validated
+     * @see https://github.com/swiftmailer/swiftmailer/issues/844
+     * @return boolean
+     */
```
fabpot: We need to ask permission to reuse this piece of code as PHPMailer is released under the GPL license, whereas we are using MIT for Swiftmailer.

lib/classes/Swift/Transport/MailTransport.php
```diff
@@ -249,7 +278,9 @@ private function _getReversePath(Swift_Mime_Message $message)
     private function _formatExtraParams($extraParams, $reversePath)
     {
         if (false !== strpos($extraParams, '-f%s')) {
-            // no need to escape $reversePath) as mail() already does it
+            if (false === $this->_isShellSafe($reversePath)) {
```
Zenexer: Unlike on line 200, there's no need to explicitly check for `false` here.  `if (!$this->_isShellSafe($reversePath)` should be sufficient.

lib/classes/Swift/Transport/MailTransport.php
Zenexer: I'm going to be releasing my own version under public domain shortly.

lib/classes/Swift/Transport/MailTransport.php
```diff
@@ -249,7 +278,9 @@ private function _getReversePath(Swift_Mime_Message $message)
...
+                $reversePath = null;
+            }
             $extraParams = empty($reversePath) ? str_replace('-f%s', '', $extraParams) : sprintf($extraParams, $reversePath);
```
fabpot: I would do:

```php
            if (empty($reversePath) || false === $this->_isShellSafe($reversePath)) {
                $extraParams = str_replace('-f%s', '', $extraParams);
            } else {
                $extraParams = sprintf($extraParams, $reversePath);
            }
```

lib/classes/Swift/Transport/MailTransport.php
barryvdh: I asked for permission by @Zenexer who is the original author. Would that not suffice?

lib/classes/Swift/Transport/MailTransport.php
fabpot: Here, it's more a coding standard than a strict requirement.

lib/classes/Swift/Transport/MailTransport.php
barryvdh: I changed this as suggested

lib/classes/Swift/Transport/MailTransport.php
barryvdh: I changed it, but do you prefer it back as `false !== ` @fabpot ?

lib/classes/Swift/Transport/MailTransport.php
```diff
@@ -239,6 +239,38 @@ private function _getReversePath(Swift_Mime_Message $message)
     }
 
     /**
+     * Fix CVE-2016-10074 by disallowing potentially unsafe shell characters.
+     *
+     * Note that escapeshellarg and escapeshellcmd are inadequate for our purposes, especially on Windows.
+     *
+     * @param string $string The string to be validated
+     *
+     * @see https://github.com/swiftmailer/swiftmailer/issues/844
```
fabpot: This one can be removed.

tests/unit/Swift/Transport/MailTransportTest.php
```diff
@@ -391,7 +418,7 @@ public function testMessageHeadersOnlyHavePHPEolsDuringSending()
     }
 
     /**
-     * @expectedException Swift_TransportException
+     * @expectedException \Swift_TransportException
```
fabpot: should be reverted, we don't have namespace in Switfmailer

lib/classes/Swift/Transport/MailTransport.php
```diff
@@ -249,8 +281,11 @@ private function _getReversePath(Swift_Mime_Message $message)
     private function _formatExtraParams($extraParams, $reversePath)
     {
         if (false !== strpos($extraParams, '-f%s')) {
-            // no need to escape $reversePath) as mail() already does it
-            $extraParams = empty($reversePath) ? str_replace('-f%s', '', $extraParams) : sprintf($extraParams, $reversePath);
+            if (empty($reversePath) || !$this->_isShellSafe($reversePath)) {
```
fabpot: Indeed, I would prefer to keep `false ===`

tests/unit/Swift/Transport/MailTransportTest.php
barryvdh: Sorry, your fabbot.io told me to, I can revert but woould probably trigger a false check on fabbot, see the patch here: http://fabbot.io/patch/swiftmailer/swiftmailer/846/aa17624da761b7a12ffa46590be1f06ed7ae9c4a/cs.diff

fabpot: I know, that's an issue with PHP CS Fixer :(

barryvdh: Ok, changed

barryvdh:
@Zenexer Can you explicitly give permission to use you code in Swiftmailer under the MIT license?

fabpot:
Thanks @barryvdh. If @Zenexer is ok for us to use his snippet of code, I will go ahead and release a new version of Swiftmailer.

Actually, I will do one more thing: deprecate the Mail transport and remove it in 6.0.

fabpot:
905b0bc deprecates the mail transport. The reasoning being that as `mail()` has many drawbacks besides security (as mentioned in the phpdocs) and because this is just a wrapper on top of `sendmail`, there is no need to keep it.

fabpot:
@barryvdh Can you add a note in the `CHANGES` file?

fabpot:
Thank you @barryvdh.

barryvdh:
Whoops, was a minute to late with my changes

Zenexer:
[This variation](https://gist.github.com/Zenexer/40d02da5e07f151adeaeeaa11af9ab36#sanitizing-input-for-safe-shell-parsing) is now in the public domain and has a few additional sanity checks to make it more generic.

It's worth noting that using sendmail directly through popen is also insecure.  The only way to securely start a process while passing user input as an argument is with pcntl_fork + pcntl_exec because it uses execve instead of the shell.

Zenexer:
Yup, this is what happens when I code at 5 AM.  Nice catch. :)

Zenexer:
Fixed.  It was meant to be `if (!$length)`.

You can use it without attribution under whatever license you like.  Just keep in mind that people can copy that portion of code without being bound by some restrictions of MIT; for example, I believe they can copy any minor modifications you make without providing attribution.

barryvdh:
> 905b0bc deprecates the mail transport. The reasoning being that as mail() has many drawbacks besides security (as mentioned in the phpdocs) and because this is just a wrapper on top of sendmail, there is no need to keep it.

@fabpot This is not the case for Windows, right? There it's a wrapper for SMTP or an alternative sendmail binary (if set in the php.ini), so not 100% drop in (eg. same could would work with MailTransport on windows + linux without application config (is php is configured to handle mail properly), but sendmail would only work on Linux right?

fabpot:
If it's a wrapper for SMTP, then the same argument applies, just use the SMTP server. The point being that whatever the configuration, mail() relies on SMTP or sendmail and we already have support for those.

PR #847, opened 2016-12-29
Clarify CVE / security fix

barryvdh:
Not sure if you want to be this explicit, but took the example from below.
Also, there is an CVE for Swiftmailer, instead those two from phpmailer, see https://legalhackers.com/advisories/SwiftMailer-Exploit-Remote-Code-Exec-CVE-2016-10074-Vuln.html

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,7 +4,11 @@ Changelog
 5.4.5 (2016-XX-XX)
 ------------------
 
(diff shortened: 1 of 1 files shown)


fabpot:
Perfect!

fabpot:
Thank you @barryvdh.

Issue #848, opened 2016-12-29
Still vulnerable


Zenexer:
https://github.com/swiftmailer/swiftmailer/blob/de19df332219d73a2704525ba75aabd7dfaa303b/lib/classes/Swift/Transport/SendmailTransport.php#L116

I think it's still vulnerable there, especially on Windows.  Am I correct in assuming the _isShellSafe check doesn't run on that at any point?  If that's the case, then it's vulnerable.

barryvdh:
So `escapeshellarg` isnt sufficient? Is that just on Windows?
https://3v4l.org/QbmV9

Zenexer:
@barryvdh It's insufficient in all but the most traditional environments, and even then, it's arguably incomplete.  Even dash will throw it off, which is used by Debian and Ubuntu.

barryvdh:
@Zenexer Zo #850?

Zenexer:
@barryvdh There are a lot of vulnerable applications and frameworks floating around, and PHP probably isn't going to patch it anytime soon.  I'd rather not provide a working PoC, even privately.  There should be sufficient information in that gist to demonstrate that abuse is possible.  You're not going to be able to make an exploit-style test case because it's technically infeasible to patch and impossible to evaluate all potential methods of abuse; you'll just need to ensure that the characters are adequately restricted in a variety of locales.

barryvdh:
But on windows, is the sendmail even vulnerable? It requires a 'fake' alternative binary, like http://glob.com.au/sendmail/ which seems to forward to SMTP or IIS pickup, so does it even support  things like the -X flag?

Zenexer:
@barryvdh I'm going to add a simplified description of the vulnerabilities to [that gist](https://gist.github.com/Zenexer/40d02da5e07f151adeaeeaa11af9ab36#unassigned) in a moment; hopefully it'll clear up some of the confusion.

Don't assume that just because -X isn't available, the vulnerability can't be abused.  You'd still be allowing the user to manipulate the command in ways that you can't predict or control, which could have any number of consequences.

PR #850, opened 2016-12-29
[Security] Use _isShellSafe in Sendmail transport also

barryvdh:
See #848
According to @Zenexer the `escapeshellarg` is not sufficient in this case. I'm not sure what payload would exactly cause this and on what systems, so following @zenexer here.

 - Created abstract Transport class to combine the reverse path + safe check
 - Make Mail transport + abstract SMTP transport extend the base abstract transport
 - Use safe check in Sendmail transport

### DIFF
--- a/lib/classes/Swift/Transport/AbstractTransport.php
+++ b/lib/classes/Swift/Transport/AbstractTransport.php
@@ -54,9 +54,18 @@ abstract class Swift_Transport_AbstractTransport
         $length = strlen($string);
         for ($i = 0; $i < $length; ++$i) {
             $c = $string[$i];
-            // All other characters have a special meaning in at least one common shell, including = and +.
-            // Full stop (.) has a special meaning in cmd.exe, but its impact should be negligible here.
-            // Note that this does permit non-Latin alphanumeric characters based on the current locale.
+
+            /*
+             * All other ASCII symbols have a special meaning in at least one common
+             * shell.  Notably, argument delimiters in cmd.exe include not only
+             * whitespace, but also comma, equals, and semicolon.  Plus also needs
+             * to be escaped in cmd.exe.
+             *
+             * ctype_alnum may allow non-Latin characters in some locales.  This
+             * should not present a problem on a properly configured system, or even
+             * most improperly configured systems.  If you need to enforce 7-bit
+             * ASCII, set the current language to "C".
+             */
             if (!ctype_alnum($c) && strpos('@_-.', $c) === false) {
                 return false;
             }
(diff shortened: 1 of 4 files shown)


lib/classes/Swift/Transport/AbstractTransport.php
```diff
@@ -0,0 +1,76 @@
...
+    {
+        $return = $message->getReturnPath();
+        $sender = $message->getSender();
+        $from = $message->getFrom();
+        $path = null;
+        if (!empty($return)) {
+            $path = $return;
+        } elseif (!empty($sender)) {
+            // Don't use array_keys
+            reset($sender); // Reset Pointer to first pos
```
barryvdh: The `_getReversePath` differ in this case, not sure which one is later?

barryvdh:
Note that this does seem to escape on Linux properly, for the give POC exploit; https://3v4l.org/QbmV9

If there are any attack vectors, I suggest to not post them here but email @fabpot directly, to avoid new public exploits.

fabpot:
I don't think that SMTP and Sendmail transports are vulnerable as they don't exec anything. The problem with the `mail()` PHP function is that it exec a command.

barryvdh:
I didn't change SMTP, but sendmail opens a process here: https://github.com/swiftmailer/swiftmailer/blob/e34123081ded3fbb4d8ccdff9ece3a4c8f6046b5/lib/classes/Swift/Transport/StreamBuffer.php#L284-L304 using the command `'/usr/sbin/sendmail -bs'` with the extra parameters here: https://github.com/swiftmailer/swiftmailer/blob/e34123081ded3fbb4d8ccdff9ece3a4c8f6046b5/lib/classes/Swift/Transport/SendmailTransport.php#L115-L119

(The SendMail extends the SMTP abstract class, but there are 2 types for the Buffer, and Sendmail is `Swift_Transport_IoBuffer::TYPE_PROCESS`)

fabpot:
Were you able to exploit this? I wasn't.

barryvdh:
No I was not, don't have a working Windows setup right now so not sure how to verify. As said I rely on  @Zenexer which can hopefully provide a working exploit (privately, not on github).

(But if `escapeshellarg` is not safe, that is probably a PHP bug)

barryvdh:
I'll close this now, can be revisited by @fabpot when there is a valid testcase.
For info see also https://gist.github.com/Zenexer/40d02da5e07f151adeaeeaa11af9ab36

sstok:
Windows uses smtp (from php.ini) not sendmail, so the the problem actually existent? 🤔

Issue #819, opened 2016-10-19
Broken pipe - unable to exit without exception


mzstic:
Hi,
we are sending quite a lot of emails. And we encountered problem with swift mailer.
What happens step by step:
- Sending mails, everything is OK
- Some problem when communicating with SMTP server
- Exception is thrown
- Exception is caught
- exit; is called
- __destruct() in Swift_Transport_AbstractSmtpTransport class is called
- it tries to call $this->stop();
- it tries to call following function $this->executeCommand("QUIT\r\n", array(221));
- it dies with exception, because it tries to write to invalid stream

Exception I'm getting is following:
Stack trace:
#0 [internal function]: {closure}(8, 'fwrite(): send ...', '/home/www/skrz....', 232, Array)
#1 vendor/swiftmailer/swiftmailer/lib/classes/Swift/Transport/StreamBuffer.php(232): fwrite(Resource id #1588, 'QUIT\r\n')
#2 vendor/swiftmailer/swiftmailer/lib/classes/Swift/ByteStream/AbstractFilterableInputStream.php(173): Swift_Transport_StreamBuffer->_commit('QUIT\r\n')
#3 vendor/swiftmailer/swiftmailer/lib/classes/Swift/ByteStream/AbstractFilterableInputStream.php(92): Swift_ByteStream_AbstractFilterableInputStream->_doWrite('QUIT\r\n')
#4 vendor/swiftmailer/swiftmailer/lib/classes/Swift/Transport/AbstractSmtpTransport.php(276): Swift_Byt in /home/www/skrz.cz/commits/57a74f2/vendor/swiftmailer/swiftmailer/lib/classes/Swift/Transport/StreamBuffer.php on line 232

Ideally I'd love to quit without getting this exception...
`try { exit; } catch ($e) {}`
I cannot even get transport and call `$transport->stop();` because it will just throw error (which I can catch and exit and be in the same situation as on the begginin) :/

PR #856, opened 2017-01-13
No exception from Swift_Transport_AbstractSmtpTransport::__destruct

mzstic:
Destructor should not throw any error because it leads to fatal error which cannot be handled.

When stream became invalid (broken pipe etc.), I was unable to terminate application without error.
And as it is stated here little bit down: http://au2.php.net/__construct
> Attempting to throw an exception from a destructor (called in the time of script termination) causes a fatal error.


mzstic:
This will resolve following issue: https://github.com/swiftmailer/swiftmailer/issues/819

fabpot:
Thank you @mzstic.

PR #860, opened 2017-01-16
update date format

rwitchell:


fabpot:
Thanks for your PR, but Swiftmailer 4 is not maintained anymore and PEAR packages are not generated anymore.

PR #864, opened 2017-01-26
Fixes an 'Undefined variable: pipes' notice.

BrunoDeBarros:
This is not a significant issue by any stretch of the imagination, but it's a quick fix and there's no reason for this notice to be there.


fabpot:
Good catch, thanks @BrunoDeBarros.

Issue #867, opened 2017-02-01
@(error control operator) at stream_socket_client


ivanproskuryakov:
added @(error control operator) character to stream_socket_client, makes impossible to debug and see the error at the stage when swiftmailer is sending a message.

"symfony/swiftmailer-bundle": "2.4.2"
https://github.com/swiftmailer/swiftmailer/blob/5.x/lib/classes/Swift/Transport/StreamBuffer.php#L267

Error sample:
```Warning: stream_socket_client(): unable to connect to smtp.mandrillapp.com:25 (Operation timed out)```

PR #868, opened 2017-02-01
@(error control operator) at stream_socket_client #867

ivanproskuryakov:
PR for https://github.com/swiftmailer/swiftmailer/issues/867

### DIFF
--- a/lib/classes/Swift/Transport/StreamBuffer.php
+++ b/lib/classes/Swift/Transport/StreamBuffer.php
@@ -264,7 +264,7 @@ class Swift_Transport_StreamBuffer extends Swift_ByteStream_AbstractFilterableIn
             $options = array_merge($options, $this->_params['stream_context_options']);
         }
         $streamContext = stream_context_create($options);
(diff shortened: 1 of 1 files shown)


lib/classes/Swift/Transport/StreamBuffer.php
```diff
@@ -264,7 +264,7 @@ private function _establishSocketConnection()
             $options = array_merge($options, $this->_params['stream_context_options']);
         }
         $streamContext = stream_context_create($options);
-        $this->_stream = @stream_socket_client($host.':'.$this->_params['port'], $errno, $errstr, $timeout, STREAM_CLIENT_CONNECT, $streamContext);
+        $this->_stream = stream_socket_client($host.':'.$this->_params['port'], $errno, $errstr, $timeout, STREAM_CLIENT_CONNECT, $streamContext);
```
fabpot: `$errno, $errstr` should contain the relevant error, and they are part of the exception message. So, I suppose you found a case where it was not the case?

fabpot:
Closing as there is no feedback.

PR #873, opened 2017-02-10
Fix tiny spelling

Rotzbua:
Just some tiny spelling.


fabpot:
Good catch, thanks @Rotzbua.

PR #874, opened 2017-02-10
add composer.lock to .gitignore

Rotzbua:
I think this is missing


fabpot:
Good catch, thanks @Rotzbua.

Issue #870, opened 2017-02-05
Swift_Signers_DKIMSigner signing algorithm


Rotzbua:
2 suggestions for improvement:

(1) Actually Swift_Signers_DKIMSigner uses `sha1` it should be changed to `sha256`.   According to rfc6376  3.3.
> Signers MUST implement and SHOULD sign using rsa-sha256.

(2) Also the code which decide which algorithm is uses should be reworked. A `switch` is used with no `default` which may cause an unset variable.

I can make an PR, you agree with the suggestions above. Some feedback please :)

PR #875, opened 2017-02-10
Update dkim

Rotzbua:
As suggested in #870.

Changes:
 - use sha256 as default
 - old php 5.3 use sha1 still as default, because sha256 is missing
 - test adjusted


fabpot:
Thank you @Rotzbua.

PR #878, opened 2017-02-13
removed usages of version_compare

fabpot:


sonnysavage:
I assume this is for performance reasons. Is that correct? Seems to me that `version_compare()` is more plain.

PR #884, opened 2017-02-15
DKIM - increase required key size

Rotzbua:
Some providers (google) ignores key sizes below 1024 bit, because they can be cracked.
So check input key size and throw error if dkim signature would be useless.


Rotzbua:
included in new #886

PR #885, opened 2017-02-18
add php 7.1 to travis

Rotzbua:
For testing but can fail.


fabpot:
Looks like tests pass, so we can move PHP 7.1 to supported.

Rotzbua:
@fabpot  I updated my commit, now php7.1 must run without errors. Btw. I think (hard) support for php versions 5.3 and 5.4 can be dropped...

fabpot:
Thank you @Rotzbua.

Issue #877, opened 2017-02-12
DKIM signature fails if body is empty


Rotzbua:
DKIM sigature is wrong if no body is set for an email. It seems that there is a rfc violation.

## Description
Default mode of class is `simple/simple`, but for body is the wrong body hash calculated, it is the body hash of mode `relaxed`. For `simple` it should be a different hash.

## Received
> DKIM-Signature: v=1; a=rsa-sha256;
> bh=47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=;

## Expected
> DKIM-Signature: v=1; a=rsa-sha256;
> bh=frcCV1k9oG9oKj3dpUqdJg1PxRT2RSN/XKdLCPjaYaY=;

## Example code
```php
    $message = Swift_Message::newInstance()
        ->setFrom(array('test@example.com' => 'example'))
        ->setTo($req['email'])
        ->setSubject('mysubject')
        ->setBody('');
```

## RFC
https://tools.ietf.org/html/rfc6376

```
3.4.3.  The "simple" Body Canonicalization Algorithm

   The "simple" body canonicalization algorithm ignores all empty lines
   at the end of the message body.  An empty line is a line of zero
   length after removal of the line terminator.  If there is no body or
   no trailing CRLF on the message body, a CRLF is added.  It makes no
   other changes to the message body.  In more formal terms, the
   "simple" body canonicalization algorithm converts "*CRLF" at the end
   of the body to a single "CRLF".

   Note that a completely empty or missing body is canonicalized as a
   single "CRLF"; that is, the canonicalized length will be 2 octets.

   The SHA-1 value (in base64) for an empty body (canonicalized to a
   "CRLF") is:

   uoq1oCgLlTqpdDX/iUbLy7J1Wic=

   The SHA-256 value is:

   frcCV1k9oG9oKj3dpUqdJg1PxRT2RSN/XKdLCPjaYaY=


3.4.4.  The "relaxed" Body Canonicalization Algorithm

   The "relaxed" body canonicalization algorithm MUST apply the
   following steps (a) and (b) in order:

   a.  Reduce whitespace:

       *  Ignore all whitespace at the end of lines.  Implementations
          MUST NOT remove the CRLF at the end of the line.
       *  Reduce all sequences of WSP within a line to a single SP
          character.

   b.  Ignore all empty lines at the end of the message body.  "Empty
       line" is defined in Section 3.4.3.  If the body is non-empty but
[...]

PR #886, opened 2017-02-18
DKIM improved

Rotzbua:
Hi,
I improved dkim a little bit.

### updated
- setters of timestamp, canon improved
- expriation timestamp was relative, but rfc says absolute and independend of signing timestamp
- ignored headers as recommended
- fore a good private key: Some providers (google) ignores key sizes below 1024 bit, because they can be cracked. So check input key size and throw error if dkim signature would be useless.

### added
- optional dkim parameter for public key query method
- dkim rfc doc

### fixes
- #877 

### Incompatible changes
- `setSignatureTimestamp()` and `setSignatureExpiration()` will not accept strings anymore.
- Keys below 1024bit can't be used anymore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 /tests/acceptance.conf.php
 /tests/smoke.conf.php
 /vendor/
+/.idea/
--- a/lib/classes/Swift/Signers/DKIMSigner.php
+++ b/lib/classes/Swift/Signers/DKIMSigner.php
@@ -51,7 +51,7 @@ class Swift_Signers_DKIMSigner implements Swift_Signers_HeaderSigner
      * 
      * @var int
      */
-    private $_hashAlgorithmOpenssl = -1;
+    protected $_hashAlgorithmOpenssl = -1;
 
     /**
      * Body canon method.
@@ -78,7 +78,7 @@ class Swift_Signers_DKIMSigner implements Swift_Signers_HeaderSigner
                                        'received' => true, // rfc6376
                                        'comments' => true, // rfc6376
                                        'keywords' => true, // rfc6376
-                                       'authentication-results' => true // best practice recommendation
+                                       'authentication-results' => true // good practice recommendation
     );
 
     /**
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -6,3 +6,4 @@
 /tests/acceptance.conf.php
 /tests/smoke.conf.php
 /vendor/
+/.idea/
```
xdecock: Not sure this should be commited ;)

lib/classes/Swift/Signers/DKIMSigner.php
```diff
@@ -328,14 +361,21 @@ public function setHashAlgorithm($hash)
...
-            $this->_bodyCanon = 'simple';
+        switch ($canon) {
+            case 'simple':
+                $this->_bodyCanon = 'simple';
+                break;
+            case 'relaxed':
+                $this->_bodyCanon = 'relaxed';
+                break;
+            default:
+                throw new Swift_SwiftException('Unable to set the body canon, must be one of simple or relaxed (%s given).', $canon);
```
xdecock: this will break compat, risky (altough, indeed, better)

notes/rfc/rfc4871.txt
```diff
@@ -3,6 +3,7 @@
 
 
 
+
```
xdecock: ? new line break?

.gitignore
Rotzbua: It is the config folder of IDE phpstorm avoids wrong commit. Don't know how often it is used...

notes/rfc/rfc4871.txt
Rotzbua: Hmm.. strange. I just saved them form official site. Seems they have changed their format.

lib/classes/Swift/Signers/DKIMSigner.php
Rotzbua: I think this has lower risk for incompatibility. `setSignatureTimestamp()` would be more difficult, because in actual code it accepts strings. In my pr it do only accept bool or int.

Rotzbua:
@xdecock thanks for your review. I added a notice to the exception. Is there a way to update the homepage? I didn't found a repository.

It seems that most people don't use risky keys anyway: https://www.unlocktheinbox.com/dkim-key-length-statistics/ .

### Incompatible changes
- `setSignatureTimestamp()` and `setSignatureExpiration()` will not accept strings anymore.
- Keys below 1024bit can't be used anymore.

.gitignore
fabpot: It should indeed be removed as this is something that needs to be configured locally instead.

Rotzbua:
pr will be split and renew.

Issue #69, opened 2011-09-02
addPart() does not inherit encoding


swiftmailer:
Consider:

$this->message->setCharset('iso-8859-1');
$message->setEncoder(Swift_Encoding::get8BitEncoding());
$message->setBody('foo æøå bar', 'text/html');
$message->addPart('foo æøå bar', 'text/plain');

The body correctly gets the right encoding and charset, however the alternative body (set by addPart) gets this:

Content-Type: text/plain; charset=utf-8
Content-Transfer-Encoding: quoted-printable

clearly ->addPart() does not correctly inherit the encoding and charset set in the $message class, and there is no other way to assign charset or encoding to the added part.

Thanks.

Original creation date: 2009-08-20T10:43:20Z
Original reporter: MichaelL
Original ticket: http://swiftmailer.lighthouseapp.com/projects/21527/tickets/107

fyrye:
A workaround for others until the issue is resolved is to manually define and attach the mime part rather than using the addPart.
```php
$message = \Swift_Message::newInstance();
$message->setEncoder(\Swift_Encoding::getBase64Encoding());
$message->setFrom('name@example.com')
    ->setTo('user@example.com')
    ->setSubject('Test')
    ->setBody('This is an html message', 'text/html');
$part = \Swift_MimePart::newInstance();
$part->setEncoder($message->getEncoder());
$part->setBody('This is a text message', 'text/plain');
$message->attach($part);
var_dump($message->toString());
```

Outputs:
```
string 'Message-ID: <4264df3ba4621e4ec0ea963f21ccca77@example.com>
Date: Thu, 14 Jan 2016 20:10:42 +0000
Subject: Test
From: name@example.com
To: user@example.com
MIME-Version: 1.0
Content-Type: multipart/alternative;
 boundary="_=_swift_v4_1452802242_e202ee1b9d03205d6d3e09b55539203f_=_"


--_=_swift_v4_1452802242_e202ee1b9d03205d6d3e09b55539203f_=_
Content-Type: text/plain; charset=utf-8
Content-Transfer-Encoding: base64

VGhpcyBpcyBhIHRleHQgbWVzc2FnZQ==

--_=_swift_v4_1452802242_e202ee1b9d03205d6d3e09b55539203f_=_
Content-Type: text/html; charset=utf-8
Content-Transfer-Encoding: base64

VGhpcyBpcyBhbiBodG1sIG1lc3NhZ2U=

[...]

Issue #500, opened 2014-07-02
Encoding is not applied to content added via $message->addPart()


mdiniz:
I'm sending a message with HTML and plain text content, and I would like to use 8bit encoding instead of quoted printable, so I do:

`$message->setEncoder(Swift_Encoding::get8BitEncoding())`

Later on I set the content:

```
$message->setBody('<span>hello</span>', 'text/html');
$message->addPart('hello', 'text/plain');
```

The email sent has the HTML part encoded as 8bit but the plain text part is still encoded as quoted printable. If I set the body with the plain text content and the HTML as a part, then only the text part is 8bit encoded.

That means, the encoding set in the message only applies to the body, not to the parts. I would expect it to apply to both of them.

I'm using Swift-5.2.1 from composer.

EDIT: this seems to be a rather old bug: https://github.com/swiftmailer/swiftmailer/issues/69
The problem didn't happen in version 3.3.2, but it happens in version 4 and 5

fabpot:
Closing as a duplicate of #69

PR #889, opened 2017-02-27
addPart() does not inherit encoding

Sapistudio:
addPart() does not inherit encoding


fabpot:
Can you add a test to avoid any future regressions? This closes #69 and #500.

fabpot:
Thank you @Sapistudio.

PR #893, opened 2017-03-21
Added suport for encrypted private keys (by a passprase) in DKIM Signer

dariogz:
## **Problem**
I noticed that when I want to use a private key, encrypted with a passprase, it was not possible to pass this passprase as a parameter, causing the **openssl_get_privatekey**() method to throw an exception.

This method accepts as an optional parameter a string passphrase
Documentation: http://php.net/manual/es/function.openssl-pkey-get-private.php

## **Added**
- passphrase attribute
- passphrase parameter as optional
- second parameter passphrase to openssl_get_privatekey()

### DIFF
--- a/lib/classes/Swift/Signers/DKIMSigner.php
+++ b/lib/classes/Swift/Signers/DKIMSigner.php
@@ -176,6 +176,7 @@ class Swift_Signers_DKIMSigner implements Swift_Signers_HeaderSigner
      * @param string $privateKey
      * @param string $domainName
      * @param string $selector
+     * @param string $passphrase
      */
     public function __construct($privateKey, $domainName, $selector, $passphrase = '')
     {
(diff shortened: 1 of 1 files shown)


dariogz:
Is there a problem with Travis and PHPUnit on PHP>7?

`PHP Fatal error:  Class 'PHPUnit_Framework_Constraint' not found in /home/travis/build/swiftmailer/swiftmailer/tests/IdenticalBinaryConstraint.php on line 8`

Rotzbua:
@dariogz Seems that a newer phpunit is used under php7x https://github.com/sebastianbergmann/phpunit/wiki/Release-Announcement-for-PHPUnit-6.0.0

fabpot:
Thank you @dariogz.

Issue #903, opened 2017-04-19
NTLMAuthenticator clobbers bcmath scale


laurinomme-helmes:
`\Swift_Transport_Esmtp_Auth_NTLMAuthenticator::getCorrectTimestamp` sets bcmath scale factor to 0, this affects bcmath operations executed later outside the component.
The scale factor should be restored before exit from the method.

fabpot:
Can you submit a pull request?

PR #904, opened 2017-04-20
NTLMAuthenticator - avoid clobbering bcmath scale #903

laurinomme-helmes:


fabpot:
Thank you @laurinomme-helmes.

Issue #894, opened 2017-03-24
Any release plan for release 6.x?


Rotzbua:
The  branch master tells a small story for release `6.0.0 (2014-XX-XX)`. Are there any plans for this release? @fabpot 
I would recommend a php version >7.x for release 6 to get rid of all deprecated and old functions and use type hinting for return values and simple types.

fabpot:
I'm working on it. Very soon now.

PR #906, opened 2017-04-21
[master] update to PHPUnit 6 and php7 only

Rotzbua:
### Changes
Update phpunit class names.
Update test for phpunit 6.
Update travis.

### Why?
Swiftmailer do not released a new major since 2013 and it do not seem that there are plans to do so #894.

I think the plan was to refactor and release it as v6. Now php 7 is released which has lot of changes which make it difficult to handle both versions php v5.6 and v7.x.

My suggestion is to drop php 5.x in master completely to avoid this complexity and use the new advantages of php 7.

### Incompatible
Both listeners seems to be not compatible to phpunit 6 yet.
```xml     
<listener class="Symfony\Bridge\PhpUnit\SymfonyTestsListener" />
<listener class="Mockery\Adapter\Phpunit\TestListener" />
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,6 @@ script: ./vendor/bin/simple-phpunit
 
 matrix:
     include:
-        - php: 5.5
-        - php: 5.6
         - php: 7.0
         - php: 7.1
         - php: hhvm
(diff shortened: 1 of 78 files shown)


fabpot:
Thank you @Rotzbua.

PR #907, opened 2017-04-29
Fix sorting MIME children when their types are equal

HansAdema:
Currently, if a message has two MIME children with the same content type, they will be added in their inverted order. In most cases, this won't matter, but for some e-mails (like X-ARF, which has two text/plain parts, one with human readable and one with YAML content), this is problematic.

This small fix ensures the original ordering is preserved when the typePrefs are identical.

### DIFF
--- a/lib/classes/Swift/Mime/SimpleMimeEntity.php
+++ b/lib/classes/Swift/Mime/SimpleMimeEntity.php
@@ -787,13 +787,13 @@ class Swift_Mime_SimpleMimeEntity implements Swift_Mime_MimeEntity
         // Sort in order of preference, if there is one
         if ($shouldSort) {
             // Group the messages by order of preference
-            $sorted = [];
+            $sorted = array();
             foreach ($this->_immediateChildren as $child) {
                 $type = $child->getContentType();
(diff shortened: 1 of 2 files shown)


fabpot:
Thanks for the fix. Can you add a test for that behavior?

HansAdema:
Sure, I will add a test for it ASAP.

HansAdema:
After adding my test, it turns out that this fix causes a regression for PHP 5.

The reason for this is the ordering of the first and second argument of `usort` being different in PHP 5 and PHP 7.

Given a simple proof of content:
```
<?php
$array = ['one', 'two'];

usort($array, function($a, $b) {
	echo 'A: '.$a.' - B: '.$b;
	return 0;
});
```

The output for PHP 5.6 is:
```
A: two - B: one
```

But for PHP 7.0, the ouput is:
```
A: one - B: two
```

In other words, you can't make any assumptions about `$a` and `$b` regarding their position in the source array.

Because of that, I rewrote the entire sorting function to not rely on `usort` anymore. This implementation groups the mime parts first, sorts those and then flattens the sorted groups. This is the only method I could think of which reliably works on all PHP versions.

fabpot:
Can you rebase? I've fixed Travis conf. Thanks

HansAdema:
Thanks, I've rebased the branch and all tests are green now.

fabpot:
Thank you @HansAdema.

PR #908, opened 2017-04-30
add phpstorm ide config to gitignore

Rotzbua:
### Reference
as mentioned in #886 https://github.com/swiftmailer/swiftmailer/pull/886/files#diff-a084b794bc0759e7a6b77810e01874f2

### Reason
Avoid accidental push oflocal config.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,15 @@
 /.php_cs.cache
 /.phpunit
-/build/*
-/composer.lock
 /phpunit.xml
+/build/*
+
+# ignore local test config
 /tests/acceptance.conf.php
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,8 +1,15 @@
...
+# ignore local test config
 /tests/acceptance.conf.php
 /tests/smoke.conf.php
+
+# ignore composer dependencies
 /vendor/
+/composer.lock
+
+# ignore phpstorm ide data
+/.idea/
```
fabpot: This should be removed. This is probably something you need in your own global .gitignore, but that does not make sense to add it locally for Swiftmailer.

PR #909, opened 2017-04-30
[travis] allow php 7.x to fail

Rotzbua:
Tavis fails because it uses a newer phpunit, until somebody fix travis script to force a older version, this patch should allow temporary fail on php7.x.


fabpot:
Closing in favor of #911

PR #910, opened 2017-04-30
Add and update dkim rfc documents

Rotzbua:
Add and update latest dkim rfc documents.
A part from big request #886.

` notes/rfc/rfc4871.txt ` yes there a new line break, idk. it's the latest version from official site.


fabpot:
Thank you @Rotzbua.

PR #928, opened 2017-05-20
Update php version

garak:


fabpot:
Good catch, thanks @garak.

Issue #689, opened 2015-11-04
TLS 2.0


phpgraphics:
Hi guys,
I need to increase security on my server so i need to use TLS 2.0. After i do that i get 
Warning: stream_socket_enable_crypto(): SSL: Connection reset by peer in swift/classes/Swift/Transport/StreamBuffer.php on line 95 Fatal error: Uncaught exception 'Swift_TransportException' with message 'Unable to connect with TLS encryption' 

Could you please tell me how i could possibly sort this?

Regards,
George

stof:
Well, swiftmailer relies on http://php.net/stream_socket_enable_crypto

So your issue looks like PHP itself does not yet support TLS 2.0 (at least in the version you are using). note that it is also related to the version of openssl you are using

xdecock:
Mmmmh, TLS 2.0 doesn't exists at all

https://fr.wikipedia.org/wiki/Transport_Layer_Security Max version is  1.3 and is in final drafts stages

Issue #598, opened 2015-04-07
swiftmailer TLSv1.1 and TLSv1.2 support


ITwrx:
Does swiftmailer currently support tlsv1.1 and tlsv1.2? If not, what is the ETA for this feature? 

Thanks

mkaatman:
PCI compliance is forcing us to turn off TLSv1.0, initially it seems like swiftmailer does not support TLSv1.2 Any confirmation on this?

stof:
Swiftmailer does not implement TLS on its own. So this depends on what your PHP version supports (which actually depends on the openssl version against which PHP is linked AFAIK)

mkaatman:
PHP 5.6.20 (cli) (built: Apr  2 2016 08:28:23)
OpenSSL 1.0.1e-fips 11 Feb 2013

Both PHP and OpenSSL support TLSv1.2

Swiftmailer starts throwing:
`[19-May-2016 08:58:01 America/Chicago] PHP Fatal error:  Uncaught exception 'Swift_TransportException' with message 'Connection could not be established with host localhost [Connection refused #111]' in ...`

I'll see if I can create a complete minimal verifiable example.

stof:
Can you paste the full message, i.e. including the stack trace ?

mkaatman:
```
[19-May-2016 08:56:02 America/Chicago] PHP Fatal error:  Uncaught exception 'Swift_TransportException' with message 'Connection could not be established with host localhost [Connection refused #111]' in /site/vendor/swiftmailer/swiftmailer/lib/classes/Swift/Transport/StreamBuffer.php:265
Stack trace:
#0 /site/vendor/swiftmailer/swiftmailer/lib/classes/Swift/Transport/StreamBuffer.php(62): Swift_Transport_StreamBuffer->_establishSocketConnection()
#1 /site/vendor/swiftmailer/swiftmailer/lib/classes/Swift/Transport/AbstractSmtpTransport.php(113): Swift_Transport_StreamBuffer->initialize(Array)
#2 /site/vendor/swiftmailer/swiftmailer/lib/classes/Swift/Mailer.php(79): Swift_Transport_AbstractSmtpTransport->start()
#3 /site/public_html/includes/mail.php(67): Swift_Mailer->send(Object(Swift_Message))
#4 /site/includes/common.php(215): in /site/vendor/swiftmailer/swiftmailer/lib/classes/Swift/Transport/StreamBuffer.php on line 265

```

[...]

PR #935, opened 2017-05-27
Allow explicit tls1.0, tls1.1, tls1.2 for startTLS

### LINKED ISSUE #933: [Bug] Swiftm. 6.x only support tls 1.0
### PHP constant change
STREAM_CRYPTO_METHOD_TLS_CLIENT changed from tls1.x to only tls1.0
Reference: https://secure.php.net/manual/de/function.stream-socket-enable-crypto.php#119122

### Line of code
https://github.com/swiftmailer/swiftmailer/blob/master/lib/classes/Swift/Transport/StreamBuffer.php#L94

### Solution
Set option to `STREAM_CRYPTO_METHOD_TLSv1_0_CLIENT|STREAM_CRYPTO_METHOD_TLSv1_1_CLIENT|STREAM_CRYPTO_METHOD_TLSv1_2_CLIENT`

Better: make it changeable than hard coded

Rotzbua:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Doc update?   | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #933 #598 #689
| License       | MIT

Swiftmailer only allow tls1.0 connections for startTLS
To support modern tls we allow explicit tls1.0, tls1.1, tls1.2


Rotzbua:
@fabpot This patch also for 6.1?

lib/classes/Swift/Transport/StreamBuffer.php
```diff
@@ -87,7 +87,11 @@ public function setParam($param, $value)
 
     public function startTLS()
     {
-        return stream_socket_enable_crypto($this->stream, true, STREAM_CRYPTO_METHOD_TLS_CLIENT);
+        // STREAM_CRYPTO_METHOD_TLS_CLIENT only allow tls1.0 connections (some php versions) 
+        // To support modern tls we allow explicit tls1.0, tls1.1, tls1.2
+        // Ssl3 and older are not allowed because they are vulnerable
+        // @TODO make tls arguments configurable
+        return stream_socket_enable_crypto($this->stream, true, STREAM_CRYPTO_METHOD_TLSv1_0_CLIENT | STREAM_CRYPTO_METHOD_TLSv1_1_CLIENT | STREAM_CRYPTO_METHOD_TLSv1_2_CLIENT);
```
fabpot: The second argument of `stream_socket_enable_crypto` is now optional. In that case, it takes whatever was configured on the stream context. So, wouldn't it be "better" to remove the second argument here, set a default on the stream when it's created and let the user configure it via `setParam()` like other stream context options?

Rotzbua: I think `stream_socket_enable_crypto` overwrites the settings of steam context, right?
So `stream_socket_enable_crypto($this->stream, true);` would be the correct code if we would use `setParms`. But then a documentation about enabeling tls1.1/1.2 for php7.0 and 7.1 should be added and thats a bigger patch and not this fast fix..

BTW with php7.2 they reverted STREAM_CRYPTO_METHOD_TLS_CLIENT to all versions of tls1.X, so the problem is only in php7.0 and 7.1

ben-crowhurst:
Hey All 👋
How can I help to get this over the line?  Is there further development required? if so spell it out and I'll get it done.

fabpot:
I messed up with this PR (or the tool I use messed up). I had to reconstrcut the commit, sorry about that. But the code is on master now.

Issue #736, opened 2016-03-16
Expose _userContentType


tremby:
I need a way to get the content type of the message body after a part has been added. After such a part has been added, `$message->getContentType()` changes to `multipart/alternative`, and even though I can get the body with `getBody()` I can't get its type.

Am I missing something, or is it impossible to get this at the moment?

It seems to me either the type should be set to `multipart/alternative` at send time rather than earlier, so a user can still get at the body and its type. Or when a part is added the existing body should immediately become a child and be accessed as such, and then the message's type can change. But perhaps there are good design decisions for the way things currently work which I'm missing.

(I posted on Stackoverflow with [a question about how to retrieve this value](http://stackoverflow.com/questions/36025109/swiftmailer-get-body-contenttype-after-a-part-has-been-added) before digging into the code and finding this.)

thomasbilk:
The information is still there in the protected property `_userContentType`. Maybe this could be solved with another method like `->getBodyContentType()` or something like that?

tremby:
That sounds good to me.

yareckon:
Made a pull request to simply return that value on ->getBodyContentType()

TerjeBr:
Any updates on this?

Why has this not been merged yet?

TerjeBr:
@fabpot or any other maintainer? This is ready to be merged.

jbondhus:
Is anyone going to merge this? I could use this as well.

TerjeBr:
@xdecock Could you merge this please?

lambjs:
@xdecock Need this too. Can you please merge?

TerjeBr:
This seems to have been merged 26th Aug. 2018 in https://github.com/swiftmailer/swiftmailer/pull/939
@fabpot could you please confirm this?

Seems like all that is left to do is to mark this PR as merged and closed.

fabpot:
Yep, fixed.

PR #939, opened 2017-05-30
Expose original body content type with getBodyContentType()

yareckon:


yareckon:
would fix #736

fmarchalemisys:
I could really use this patch right now. Any reason why such a simple patch has not been merged yet?

yareckon:
ping :)

LucWollants:
What is the status on this? I would like to use this method in a unit test. I have a html and plain text alternative and wan't to make sure both are present.

tremby:
This patch seems to add some trailing whitespace to a couple of lines (170, 180). I think if that were the reason it would have been said by now, but maybe worth tidying it up just in case?

Issue #948, opened 2017-06-06
Where is Swift_Validate in 6.*?


Laurent3170:
The file `swiftmailer/swiftmailer/lib/classes/Swift/Validate.php` and its class `Swift_Validate`has disappeared from v6.* but still mentionned in https://github.com/swiftmailer/swiftmailer/blob/master/doc/messages.rst and NOT appearing has deleted in https://github.com/swiftmailer/swiftmailer/blob/master/CHANGES.
So, is it a mistake? If not what is the new way to validate?

Laurent3170:
So... If I want to validate, I must stick to 5.4.8?

Laurent3170:
@fabpot sorry to ping you, but it seems to me that having lost Validation is A problem. So if you have any answer...

Rotzbua:
@Laurent3170 Stop crying, use the search then you will find the composer file or the according commit https://github.com/swiftmailer/swiftmailer/commit/6776d293954416a9ee7df4d76f24cea6eb3370a9 and then you see that it has been replaced by an external library.
Then make a pull request which adds this change to the changes.txt and update the doc, that helps fabpot and all other user. Thanks.

Laurent3170:
@Rotzbua am I crying? No I don't think so... But if so, thanks for having drying my tears...
Maybe you're a geek in Github and Symfony and maybe I'm not.
I have used the searched and just found that the file was not there anymore, with no explanation, looking like it was just a mistake. Don't you do some mistakes as I do? First question was 
> So, is it a mistake?

Not having this answer gives no possibility to know. I can't re-do all the way that has lead to this decision if I was not implied in these discussions.
Second question was 
> If not what is the new way to validate?

You gave me an answer with the commit url and I'll look at it, and I thank you for this part of your answer.
And then I will be able to do a commit, if still needed, as you have mentioned and as I did in other parts of other bundles.

Laurent3170:
@Rotzbua I followed your link and created two PR's #951 and #952 (yes I don't know how to do only one directly from Github, but ready to learn).

fabpot:
@Laurent3170 You did great! Thank you.

PR #951, opened 2017-06-12
Add of new way to validate email

Laurent3170:
According to issue https://github.com/swiftmailer/swiftmailer/issues/948, explain the new way to validate.


fabpot:
Thank you @Laurent3170.

PR #955, opened 2017-06-15
openssl_random_pseudo_bytes is not used anymore

Rotzbua:
Just a code cleanup.
openssl_random_pseudo_bytes was replaced earlier by random_bytes (php7)


fabpot:
Good catch, thanks @Rotzbua.

Rotzbua:
@fettymkima I guess you posted your comment in the wrong place..

Issue #956, opened 2017-06-16
SpoolTransport method send shouldn't return the boolean instead of integer ?


apodgorbunschih:
https://github.com/swiftmailer/swiftmailer/blob/master/lib/classes/Swift/Transport/SpoolTransport.php

The method send, shouldn't return boolean instead of integer? and also the phpdoc says that it returns "The number of sent e-mail's" but in reality it returns or 1 or 0.

Rotzbua:
Please notify the `implements Swift_Transport` and the interface mention `int` as return type.

PR #958, opened 2017-06-19
migrate to php7 Swift_Transport

Rotzbua:
- add return types
- replace redundant documentation

Reference to #956


fabpot:
I'm closing this PR as we won't migrate to type-hints. Swiftmailer will be replace by Symfony Mailer soon.

PR #959, opened 2017-06-21
[doc] add link to stack overflow

Rotzbua:


fabpot:
Good catch, thanks @Rotzbua.

Issue #657, opened 2015-08-27
Exclude unnecessary files from git archive


fbm-static:
Hello,

Could it be considered to add `export-ignore`  to files not required for production when installing or requiring this package to composer?

Example of these are:
- tests
- docs
  - phpunit.xml.dist
  - .travis.yml
  - CHANGES
  - README

Thank you.

replete:
i++;

fabpot:
Already discussed (on some of my other projects) and rejected. See the discussion in symfony/symfony#17749 for instance.

PR #963, opened 2017-06-29
Ignore tests suite on git export

guiwoda:
This prevents composer downloading the lib's tests suite on dependent projects.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -7,3 +7,4 @@
 
 # ignore /notes in the git-generated distributed .zip archive
 /doc/notes export-ignore
+/tests export-ignore


Rotzbua:
According to #657 ?

guiwoda:
@Rotzbua exactly. Other files can be added as well, so that they are not downloaded through composer.

fabpot:
Already discussed (on some of my other projects) and rejected. See the discussion in symfony/symfony#17749 for instance.

guiwoda:
@fabpot The issue you point is not so much a discussion as a decision with some arguments, but without responding to the counter arguments. I agree with the counter arguments. 

I think it's a shame you guys are forcing down library test code on our production servers and making us do the heavy lifting of filtering it out, which by the way I'm not even sure how to do unless I make a script that does specific clean-up for each dependency and transient dependency. 

Hope you revisit this issue and the decision you took. I'd be glad to participate in the debate if and when it happens.

PR #983, opened 2017-08-09
replace usage of getmypid()

### LINKED ISSUE #745: getmypid() has been disabled for security reasons
hi ! 

I was looking at #445 but I still have a problem with this function.
When I look at the sources, this function is still used 3 times in https://github.com/swiftmailer/swiftmailer/blob/5.x/lib/classes/Swift/Mime/SimpleMimeEntity.php

Could this be fixed or am I doing it wrong ?

Rotzbua:
* with php7 there is a good way to get random strings with
random_bytes()
* length not changed
* fix #745

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,7 @@ Changelog
 6.0.2 (2017-XX-XX)
 ------------------
 
+ * removed usage of getmypid()
  * n/a
 
 6.0.1 (2017-05-20)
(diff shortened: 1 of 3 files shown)


Rotzbua:
Notice: there is no change in length like in a first attempt which broke the id. See 84ff290871c3436c3703ce15ec5e344da0b59472

fabpot:
Thank you @Rotzbua.

houyimin:
I got a error with this, PHP Fatal error:  Call to undefined function random_bytes() in \swiftmailer\swiftmailer\lib\classes\Swift\Mime\SimpleMimeEntity
.php on line 424,is now only support PHP7?

Rotzbua:
@houyimin How about reading readme or composer.json?

houyimin:
@Rotzbua  OK,Thank you

Issue #967, opened 2017-07-18
DecoratorPlugin: PHP Recoverable fatal error:  Object of class DateTimeImmutable could not be converted to string


SkunkyDu52:
Hi,
I've got an issue with the decoration of a date. I dont know if it is intentionnal or not (that i could understand because of local date  format...)
In fact, i try to decorate an element whose value is an instance of DateTime. This snippet generate the error:

```
foreach ($message->getHeaders()->getAll() as $header) {
          $body = $header->getFieldBodyModel();
                $count = 0;
                if (is_array($body)) {
                    $bodyReplaced = array();
                    foreach ($body as $key => $value) {
                        $count1 = 0;
                        $count2 = 0;
                        $key = is_string($key) ? str_replace($search, $replace,
                                        $key, $count1) : $key;
                        $value = is_string($value) ? str_replace($search,
                                        $replace, $value, $count2) : $value;
                        $bodyReplaced[$key] = $value;

                        if (!$count && ($count1 || $count2)) {
                            $count = 1;
                        }
                    }
                } else {
                    $bodyReplaced = str_replace($search, $replace, $body, $count);
                }
```

In my case, $body isnt an array then pass in the str_replace. As it is not a string but an object, i got the following error:
 PHP Recoverable fatal error:  Object of class DateTimeImmutable could not be converted to string in ...\vendor\swiftmailer\swiftmailer\lib\classes\Swift\Plugins\DecoratorPlugin.php

Will you do something about that? I corrected it like that:
```
foreach ($message->getHeaders()->getAll() as $header) {
                $body = $header->getFieldBodyModel();
                if ($body instanceof \DateTimeImmutable) {
                    $body = $body->format('Y-m-d H:i');
                }
```

But, as you can expect, i dislike to change things in vendor...
Thanks in advance for you repply,

Maxime

HanXHX:
I confirm this issue. PHP version: 7.0.19 and and swiftmailer: 6.0.1


```php
        $replacements = [
            'blah@bloh.com' => [
                '{firstname}' => 'Foo',
                '{link}' => 'http://bar.com'
            ]
        ];

        $emailer->registerPlugin( new \Swift_Plugins_DecoratorPlugin($replacements));
        $emailer->send($message); // crash
```

pbarney:
I'm having the same problem. Has anyone come up with a solution?

[...]

Issue #957, opened 2017-06-19
Using DecoratorPlugin throws error in v6.0.0


juanped:
While trying to use the Decorator plugin below error is thrown;

`Recoverable fatal error: Object of class DateTimeImmutable could not be converted to string in ..\swiftmailer\lib\classes\Swift\Plugins\DecoratorPlugin.php on line 115`

pbarney:
This is also referenced at #967

PR #1007, opened 2017-09-30
A possible fix for the issues #957 and #967

zerkms:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc update?   | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #957 #967
| License       | MIT

The proposed fix simply checks whether the header is of a string type, since otherwise it makes very little sense to run substitutions.


fabpot:
Thank you @zerkms.

PR #1023, opened 2017-11-24
Refactored the SMime signer

### LINKED ISSUE #822: Settings Content-Transfer-Encoding multiple times during S/MIME encryption
I have a multipart message with a simple plain text part and an attachment, which is base64 encoded.

In the SMimeSigner::signMessage() the encoder for raw encoding is added.

The resulting headers contain three "Content-Transfer-Encoding" entries:

```
Content-Type: application/x-pkcs7-mime; smime-type=enveloped-data;
 name="smime.p7m"; charset=utf-8
Content-Transfer-Encoding: raw
Content-Transfer-Encoding: raw
Content-Transfer-Encoding: base64
```

I could not find anything prohibited about the three time usage of "Content-Transfer-Encoding", but it says `Unlike Content-Types, a proliferation of Content-Transfer-Encoding values is undesirable and unnecessary.` (http://www.freesoft.org/CIE/RFC/1521/5.htm)
Also that might be the reason PXI firewalls with deep inspection do reject sending those emails (not proven to be a valid statement, but the only thing we could come up with, since the unencrypted email had only one per part).

Is this a known bug or maybe even intended behavior?

### LINKED ISSUE #834: Sending encrypted email with attachment
Sending encrypted email seems to break when you try to add attachments.

I created my key as follows.
`openssl req -x509 -newkey rsa:4096 -keyout key.pem -out cert.pem -days 365`

I can send an encrypted email  without any attachments just fine.  I use the following code for the encryption part.

```
$smimeSigner = Swift_Signers_SMimeSigner::newInstance();
$smimeSigner->setEncryptCertificate('cert.pem', OPENSSL_CIPHER_3DES);
$emailMessage->attachSigner($smimeSigner);
```

I can then download the smime.pm7 and verify it worked with the following command.
`openssl smime -in test3.p7m -inform DER  -inkey key.pem -pk7out`

However, when I attach something using the following
`->attach(Swift_Attachment::fromPath('test.txt'));`

OpenSSL gives me an error when trying to open the p7m file: `Error reading S/MIME message`.

There seems to be an issue with the way the message is encrypted when it's a multi-part message.

Some servers will just accept the email, regardless that it's malformed.  However, it appears some mail servers won't respond to the `\r\n.\r\n` command if your p7m attachment can not be parsed.  So if anybody is getting a "Server timed out" or "Expected code 250" this might be the reason why.  I binded the stream to a filestream to test the output.  The last bit of output is `0d0a0d0a0d0a2e0d0a515549540d0a`.  After `0d0a2e0d0a` (\r\n.\r\n) is sent the stream times out waiting for a response and then`515549540d0a` (quit) is sent because of the destructor.

janflora:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Doc update?   | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #834, #822
| License       | MIT

* Support for full MIME message wrapping (message/rfc822) for both signing and encryption was added. This has been described in RFC5751 section 3.1.
* The methods that carries out signing and encryption have been completely rewritten and a lot of comments were added to explain what's going on. This was not very clear before.
* The above mentioned rewrites fixed #834 and fixed #822.
* A public method was added to the Swift_Message class to clear all attached signers. This way it is possible to reuse the existing message object when signing and/or encrypting.
* A "Null" MIME content encoder was added to help recreating MIME parts from byte streams. OpenSSL is used carry out the actual encryption and signing. The resulting MIME part is returned as a byte stream (file) and needs to be parsed back into a Swift message. Since the content of the byte stream already has the correct encoding, the appropriate content encoder just needs to set the correct header and not touch the body content at all.

### DIFF
--- a/lib/classes/Swift/Message.php
+++ b/lib/classes/Swift/Message.php
@@ -129,8 +129,8 @@ class Swift_Message extends Swift_Mime_SimpleMessage
      */
     public function clearSigners()
     {
-        $this->headerSigners = [];
-        $this->bodySigners = [];
+        $this->headerSigners = array();
+        $this->bodySigners = array();
 
         return $this;
     }
@@ -218,7 +218,7 @@ class Swift_Message extends Swift_Mime_SimpleMessage
         $this->savedMessage = array('headers' => array());
         $this->savedMessage['body'] = $this->getBody();
         $this->savedMessage['children'] = $this->getChildren();
-        if (count($this->savedMessage['children']) > 0 && $this->getBody() != '') {
+        if (count($this->savedMessage['children']) > 0 && '' != $this->getBody()) {
             $this->setChildren(array_merge(array($this->becomeMimePart()), $this->savedMessage['children']));
             $this->setBody('');
         }
--- a/lib/classes/Swift/Message.php
+++ b/lib/classes/Swift/Message.php
@@ -123,6 +123,19 @@ class Swift_Message extends Swift_Mime_SimpleMessage
     }
 
     /**
+     * Clear all signature handlers attached to the message.
+     *
+     * @return $this
+     */
+    public function clearSigners()
+    {
+        $this->headerSigners = [];
+        $this->bodySigners = [];
+
+        return $this;
+    }
+
+    /**
      * Get this message as a complete string.
      *
      * @return string
--- a/lib/classes/Swift/Mime/ContentEncoder/NullContentEncoder.php
+++ b/lib/classes/Swift/Mime/ContentEncoder/NullContentEncoder.php
@@ -1,0 +1,81 @@
+<?php
+
+/*
+ * This file is part of SwiftMailer.
+ * (c) 2004-2009 Chris Corbyn
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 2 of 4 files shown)


Rotzbua:
@janflora Hi, can you add some comments to the code which reference to the according rfc? It is a little bit hard to review the code if I always have to search the definition ;) Thanks.

janflora:
@Rotzbua I already did here: https://github.com/swiftmailer/swiftmailer/pull/1023/files#diff-459e6d445d75ee83376d20ec2381b158R183
Right now it goes in the description of the method that toggles the functionality. Should I move it somewhere else?

janflora:
@Rotzbua Any news on this? I'll be happy to explain the changes I did in even more detail if that will help with the code review.

janflora:
@Rotzbua @fabpot Could you please let me know what you want me to do in order to get this change merged? I actually think that I made a valuable contribution and I would really like to help out others by getting these changes into the master branch. If you let too much time pass here, I'll forget the details of the implementation and it will take me a lot of time to get back into it if changes has to be made. Thanks.

fabpot:
Thank you @janflora.

Issue #415, opened 2014-01-13
Not working with mailcatcher.me


RSully:
I was directed towards mailcatcher from laravel/framework#1821

When I set Laravel to use smtp with localhost and port 1025 swift throws the following error:

> [Swift_TransportException]
> Expected response code 220 but got code "500", with message "500 Unknown command"

MageParts:
Thanks @RSully for saving a few hours of my life :-)

GokulSrinivas:
Thanks @RSully . You saved my time :+1:

twify93:
Thank you @RSully

falnyr:
@RSully LIFESAVER!

fabpot:
Can anyone send a pull request to add a small note about how to use mailcatcher with Swiftmailer. That would help many other users.

PR #1028, opened 2017-12-11
Update sending.rst - Note using Mailcatcher

twify93:
Referencing change to the documentation regarding the Mailcatcher #415 , since it does not support encryption


fabpot:
Thank you @twify93.

PR #1034, opened 2017-12-26
add suggestions to composer

Rotzbua:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc update?   | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | no
| License       | MIT

If installed by composer, show required extensions for optional features.


composer.json
```diff
@@ -22,6 +22,12 @@
         "mockery/mockery": "~0.9.1",
         "symfony/phpunit-bridge": "~3.3@dev"
     },
+    "suggest": {
+        "ext-openssl": "required for encrypted connections, dkim, smime, NTLMAuthenticator, ...",
+        "ext-bcmath": "required for NTLMAuthenticator",
+        "ext-mbstring": "mbstring or iconv required for multi byte support",
```
fabpot: I suggest that we require `symfony/polyfill-mbstring` instead.

CHANGES
```diff
@@ -5,6 +5,7 @@ Changelog
 ------------------
 
  * n/a
```
fabpot: should be removed

composer.json
```diff
@@ -22,6 +22,12 @@
...
+        "ext-iconv": "mbstring or iconv required for multi byte support, iconv is required for NTLMAuthenticator"
```
fabpot: Same here, what about requiring `symfony/polyfill-iconv` (which is needed for the mbstring polyfill)?

fabpot (commented):
/cc @nicolas-grekas

composer.json
Rotzbua: I guess a polyfill is not required.
I think there should be no real php7.x environment without extensions mbstring or iconv.
Also "suggest" only creates a notice if the extension is not installed and will be shown after `$ composer require "swiftmailer/swiftmailer:^6.0"`. I only added the "suggestions" to create a hint for misconfigured environments while using composer and not while runtime.

fabpot: For the record, I think that "suggest" are useless. Nobody reads them. And it's a one time thing.

fabpot:

PR #1047, opened 2018-01-25
Add Swift_DependencyContainer::asArray()

c960657:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc update?   | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 
| License       | MIT

The constructor for `Swift_Transport_EsmtpTransport` accepts an array of `Swift_Transport_EsmtpHandler` instances. In order to add more handlers using `Swift_DependencyContainer` you have to specify all handlers in addition to the four other constructor arguments.

This PR makes it possible to extract the array of instances as a separate item in the dependency container.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -5,6 +5,7 @@ Changelog
 ------------------
 
  * added support for IDN domains in email addresses
+ * introduced new dependencies: transport.smtphandlers and transport.authhandlers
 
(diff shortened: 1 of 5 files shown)


fabpot:
Thank you @c960657.

PR #1049, opened 2018-01-25
Remove getPriorityOver()

c960657:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc update?   | no
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | 
| License       | MIT

The idea behind `Swift_Transport_EsmtpHandler::getPriorityOver()` is flawed.

`uasort()` requires that operators are symmetric, i.e. if `FooHandler::getPriorityOver("BAR")` returns 1, `BarHandler::getPriorityOver("FOO")` should return -1. A handler cannot simply return 0 to indicate that it does not care about ordering, like `Swift_Transport_Esmtp_AuthHandler` currently does. It is not practical for a handler to know about all other possible handlers and their priority requirements.

You usually only have a few extension handlers that are configured manually, so having an automated priority sort is not really required. I suggest that we just require users to manually sort the list of extension handlers before passing it to `Swift_Transport_EsmtpTransport::setExtensionHandlers()` or the `Swift_Transport_EsmtpTransport` constructor.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+7.0.x (2018-XX-XX)
+------------------
+
+ * removed Swift_Transport_EsmtpHandler::getPriorityOver()
+
+
 6.1.0 (2018-XX-XX)
 ------------------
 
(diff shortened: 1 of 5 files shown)


fabpot:

PR #1044, opened 2018-01-22
Basic IDN support

c960657:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc update?   | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #63, #541
| License       | MIT

IDN domains have been around for more than a decade and are now widely used. Swiftmailer should support internationalised email addresses. Even if you don't use an IDN domain yourself, you may need to send email to people who do.

Calling `setTo()` with a non-ASCII email address does not throw an error (it does throw, if the email address is otherwise malformed). The [egulias/EmailValidator](https://github.com/egulias/EmailValidator) package accepts addresses containing non-ASCII characters both in the local-part (the part of the address left of @) and domain (after @). However, when you try to send the email, the SMTP server rejects the address for being syntactically invalid.

Handling non-ASCII characters in local-part and domain are two different issues. Non-ASCII characters in the domain is supported by all email servers, as long as the domain is IDN-encoded by the sender. Non-ASCII characters in local-part requires support for the `SMTPUTF8` SMTP extension (see [RFC 6531](https://tools.ietf.org/html/rfc6531)).

This PR adds basic support for email addresses using IDN domains. It does not add support for `SMTPUTF8` but simply IDN-encodes the domain, so mail can use the existing infrastructure. The implementation is backwards-compatible fashion.

Future work should add support for `SMTPUTF8` in order to support for email addresses with non-ASCII characters in local-part also. This would probably require BC-breaking changes, so it needs to go into a new major release of Swiftmailer.

PR #1053, opened 2018-01-28
Handle address encoder exceptions during send

c960657:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc update?   | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 
| License       | MIT

After working some more with #1044 I noticed some inconsistencies between `SendmailTransport` and `EsmtpTransport`. First of all, I failed to add the address encoder to the constructor of `SendmailTransport`. And secondly, a `Swift_AddressEncoderException` throw in `doRcptToCommand()` because of a bad _recipient_ address was not caught and added to `$failedRecipients`. If the exception is thrown because of a bad sender address, we still throw the exception – I assume this is the expected behaviour.

I modified the other tests in `AbstractSmtpTest` to catch a Swiftmailer-specific exception, so they wont accidentally catch an exception thrown by PHPUnit in the `fail()` method.


fabpot:
Thank you @c960657.

PR #1056, opened 2018-02-05
Support SMTPUTF8 extension

c960657:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc update?   | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |
| License       | MIT

#1044 added support for non-ASCII characters in the email domain. This PR adds support for the SMTPUTF8 SMTP extension (see [RFC 6531](https://tools.ietf.org/html/rfc6531)) that allows UTF-8 characters in the local-part also.

This PR does _not_ implement an UTF-8 header encoder (e.g. `Swift_Mime_HeaderEncoder_Utf8HeaderEncoder`) as described in [RFC 6532](https://tools.ietf.org/html/rfc6532). For the purpose of sending to non-ASCII email addresses, the new `Swift_AddressEncoder_Utf8AddressEncoder` works fine with the existing header encoders – `Swift_Mime_Headers_MailboxHeader` does not encode the actual email address but simple passes through whatever is returned by the address encoder.

FWIW, SMTPUTF8 is supported by [Postfix](http://www.postfix.org/SMTPUTF8_README.html), [Gmail](https://googleblog.blogspot.dk/2014/08/a-first-step-toward-more-global-email.html), among others.

### DIFF
--- a/lib/classes/Swift/AddressEncoder/Utf8AddressEncoder.php
+++ b/lib/classes/Swift/AddressEncoder/Utf8AddressEncoder.php
@@ -1,0 +1,36 @@
+<?php
+
+/*
+ * This file is part of SwiftMailer.
+ * (c) 2018 Christian Schmidt
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/**
+ * A UTF-8 email address encoder.
+ *
+ * Returns the email address verbatimly in UTF-8 as permitted by RFC 6531 and
+ * RFC 6532. It supports addresses containing non-ASCII characters in both
+ * local-part and domain (i.e. on both sides of @).
+ *
+ * This encoder must be used together with Swift_Transport_Esmtp_SmtpUtf8Handler
+ * and requires that the outbound SMTP server supports the SMTPUTF8 extension.
+ *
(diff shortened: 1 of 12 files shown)


lib/classes/Swift/AddressEncoder/Utf8AddressEncoder.php
```diff
@@ -0,0 +1,36 @@
...
+ * @author Christian Schmidt
+ */
+class Swift_AddressEncoder_Utf8AddressEncoder implements Swift_AddressEncoder
+{
+    /**
+     * Returns the address verbatimly.
+     */
+    public function encodeString(string $address): string
+    {
+        return $address;
```
fabpot: Do we need to check that the address is actually proper UTF-8 and throw an exception if not?

c960657: It seems the EmailValidator will reject addresses with invalid UTF-8 encoding:
https://github.com/egulias/EmailValidator/blob/master/EmailValidator/EmailLexer.php#L208

c960657:
I have added some documentation.

fabpot:
Thank you @c960657.

PR #1059, opened 2018-02-26
Direct SMTP Transport

c960657:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc update?   | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 
| License       | MIT

In some hosting scenarious you don't have access to an outbound SMTP server. With few changes, Swift Mailer can deliver the message directly to the recipient's mail server.

This may also be useful if you need instant feedback on delivery errors, rather than waiting days for a delivery status notification to be returned.

### DIFF
--- a/lib/classes/Swift/Transport/AbstractSmtpTransport.php
+++ b/lib/classes/Swift/Transport/AbstractSmtpTransport.php
@@ -177,11 +177,6 @@ abstract class Swift_Transport_AbstractSmtpTransport implements Swift_Transport
      */
     public function send(Swift_Mime_SimpleMessage $message, &$failedRecipients = null)
     {
-        if (!$this->isStarted()) {
-            $this->start();
-        }
-
-        $sent = 0;
         $failedRecipients = (array) $failedRecipients;
 
         if ($evt = $this->eventDispatcher->createSendEvent($this, $message)) {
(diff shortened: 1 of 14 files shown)


c960657:
About the restriction that all recipients should belong to the same domain:
I guess most messages have only one recipient, so the limitation wont affect the common use-case.

I also made [another implementation](https://github.com/swiftmailer/swiftmailer/compare/027e4d7...c960657:direct-esmtp-transport) without this restriction. It was a bit more complex and required a new public methods in `Swift_Transport_AbstractSmtpTransport`, so it felt a bit clunky, but it was definitely possible.

On second thought it might be worth the effort, so the transport supports the same base functionality as the others. Any thoughts on that?

fabpot:
If we can make it work like other transports, that would be even better and would avoid some WTFs.

c960657:
I have now pushed the “advanced” implementation that allows messages to multiple destinations.

lib/classes/Swift/Transport/AbstractSmtpTransport.php
```diff
@@ -395,27 +413,6 @@ protected function streamMessage(Swift_Mime_SimpleMessage $message)
         $this->executeCommand("\r\n.\r\n", [250]);
     }
 
-    /** Determine the best-use reverse path for this message */
-    protected function getReversePath(Swift_Mime_SimpleMessage $message)
```
fabpot: As this one is protected, it should be deprecated and just modified to use the new method.

lib/classes/Swift/Transport/AbstractSmtpTransport.php
```diff
@@ -223,7 +214,34 @@ public function send(Swift_Mime_SimpleMessage $message, &$failedRecipients = nul
...
+
+        $message->setBcc([]);
+
+        try {
+            $sent += $this->sendTo($message, $reversePath, $tos, $failedRecipients);
+            $sent += $this->sendBcc($message, $reversePath, $bcc, $failedRecipients);
+        } catch (Exception $e) {
+            $message->setBcc($bcc);
+            throw $e;
+        }
```
fabpot: Can we use a `finally` clause there to avoid the duplication of the `setBcc` call?

fabpot (changes requested):

lib/classes/Swift/Transport/AbstractSmtpTransport.php
c960657: The new method is also protected, so it cannot be called from here. Instead I just restored the existing code and marked it as deprecated.

lib/classes/Swift/Transport/AbstractSmtpTransport.php
c960657: Done, thanks.

fabpot:
Can you rebase to get rid of the merge commit?

c960657:
Done.

fabpot:

PR #1064, opened 2018-03-03
Add 8BITMIME support

c960657:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc update?   | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |
| License       | MIT

It is not permitted to sent 8-bit content over plain SMTP. The [8BITMIME](https://tools.ietf.org/html/rfc6152) SMTP extension makes this possible.

This PR adds a new SMTP handler that adds support for 8BITMIME. This handler should be used e.g. when using `Swift_Mime_ContentEncoder_PlainContentEncoder` in `8bit` mode.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -8,6 +8,7 @@ Changelog
  * introduced new dependencies: transport.smtphandlers and transport.authhandlers
  * deprecated Swift_Signers_OpenDKIMSigner; use Swift_Signers_DKIMSigner instead
  * added support for SMTP pipelining
+ * added Swift_Transport_Esmtp_EightBitMimeHandler
 
 6.0.2 (2017-09-30)
 ------------------
--- a/doc/messages.rst
+++ b/doc/messages.rst
@@ -839,7 +839,7 @@ sent to this address.
 Setting the Character Set
 -------------------------
 
-The character set of the message (and it's MIME parts) is set with the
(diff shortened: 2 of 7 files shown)


fabpot:
Looks interesting. Can you add docs for this new feature?

c960657:
I have added some documentation.

Also, I introduced a new alias, `mime.textcontentencoder`, to make it easier to switch the encoder.

I have added `transport.8bitmimehandler` as default SMTP handlers. I must admit that I don't know the full impact of adding this (and `transport.smtputf8handler`). Not all SMTP servers support transcoding, if a message is received from Swift Mailer using these extensions and should be delivered to a downstream server that does not support these. However, as long as the email itself does not contain 8-bit characters in the body and address, respectively, the server should be able to just ignore the presence of these extensions. But I don't know if some servers might simply reject the message (the RFC allows that).

So I guess the conservative position for a library such as Swift Mailer should be to omit these from the default configuration, so people can enable them explicitly, if their outbound SMTP server supports them including downgrading. Do you agree? In that case, I'll remove them.

fabpot:
I agree that we should be on the conservative side of things.

c960657:
Done.

fabpot:
Thank you @c960657.

PR #1065, opened 2018-03-03
Support PIPELINING SMTP extension

c960657:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc update?   | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |
| License       | MIT

Add support for the PIPELINING SMTP extension defined in [RFC 2920](https://tools.ietf.org/html/rfc2920). This speeds up delivery, especially when using a SMTP server on a high-latency connection.

### DIFF
--- a/lib/classes/Swift/Transport/AbstractSmtpTransport.php
+++ b/lib/classes/Swift/Transport/AbstractSmtpTransport.php
@@ -30,7 +30,7 @@ abstract class Swift_Transport_AbstractSmtpTransport implements Swift_Transport
     protected $addressEncoder;
 
     /** Whether the PIPELINING SMTP extension is enabled (RFC 2920) */
-    protected $pipelining = false;
+    protected $pipelining = null;
 
     /** The pipelined commands waiting for response */
     protected $pipeline = [];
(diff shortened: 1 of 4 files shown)


lib/classes/Swift/Transport/AbstractSmtpTransport.php
```diff
@@ -311,18 +317,28 @@ public function getBuffer()
...
         $failures = (array) $failures;
         $seq = $this->buffer->write($command);
-        $response = $this->getFullResponse($seq);
         if ($evt = $this->eventDispatcher->createCommandEvent($this, $command, $codes)) {
             $this->eventDispatcher->dispatchEvent($evt, 'commandSent');
         }
-        $this->assertResponseCode($response, $codes);
+
+        $this->pipeline[] = [$command, $seq, $codes];
+        if ($pipeline && $this->pipelining) {
```
fabpot: Is there a case where you would want to not use pipelining if the SMTP supports it?

c960657: Hmm, there are some (old) reports about compatibility issues:
* http://postfix.1071664.n5.nabble.com/Disable-outbound-smtp-pipelining-td40604.html
* ftp://ftp.cs.uu.nl/mirror/postfix/workarounds.html

I have added a method that allows forcing pipelining to be enabled or disabled.

fabpot:
Thank you @c960657.

Issue #640, opened 2015-07-22
DKIM header not seperated from Message-ID


Freubert:
There is an issue where the header fields "DKIM-Signature" and "Message-ID" are mixed up into one single header which results into the following:

```
DKIM-Signature: v=1; a=rsa-sha1; c=relaxed/relaxed; d=domain.com; s=mail;
    t=1437574424; i=@domain.com; bh=2jmj7l5rSw0yVb/vlWAYkK/YBwk=;
    h=Message-ID:Date:Subject:From:To:MIME-Version:Content-Type;
    b=rXhMit3rNzhNU8qVsEYQtCWqIzbHdAdrHhLQtBFeCuODmLqHNUQkv52XUWkT8C/AqFh3
     ODnYHMXzVFmJaVh9lIXdf66l4oCx85lpdDnP7dowwk38GWjn/h1VphgkNN1lkZl3JFvfG
     6gid0inPDPhNOiDf4kY5MwFdnZDT2c/F0g=Message-ID: <a0050431c8aefd6004ae66bfefa5ce13@domain.com>
```

There's simply a line break missing.

My setup is as follows:

``` php
$mailer = Swift_Mailer::newInstance(Swift_SmtpTransport::newInstance('localhost', 25));
$signer = new Swift_Signers_OpenDKIMSigner(
    file_get_contents('/path/to/private.certificate.pem'),
    'domain.com',
    'mail');
$signer->setBodyCanon('relaxed')
    ->setHeaderCanon('relaxed');
$message = Swift_Message::newInstance($subject)
    ->setFrom(array('noreply@domain.com' => 'domain'))
    ->setTo($email, $nick)
    ->setBody($mailview->render('email.phtml'), 'text/html')
    ->addPart($mailview->render('email-text.phtml'), 'text/plain')
    ->attachSigner($signer);
return $mailer->send($message);
```

I've created a workaround where I modified Swift_Signers_OpenDKIMSigner::addSignature() in line 52 from

``` php
$header->setValue($headerVal);
```

to

``` php
$header->setValue($headerVal . PHP_EOL);
```

which resolves the issue but is quite quick & dirty.

PR #1066, opened 2018-03-14
OpenDKIM fixes

c960657:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc update?   | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #640
| License       | MIT

`Swift_Signers_OpenDKIMSigner` is severely broken. It depends on a PHP extension that is poorly documented and not included in any OS distribution (AFAIK). I suggest removing it altogether in favour of `Swift_Signers_DKIMSigner`, but FWIW here are some fixes that are necessary to make it run at all (at with least the latest version of the extension).

I didn't find a way to use `relaxed` for body — this throws an error `OpenDKIM Error: private key load failure` no matter what I do.

### DIFF
--- a/lib/classes/Swift/Signers/OpenDKIMSigner.php
+++ b/lib/classes/Swift/Signers/OpenDKIMSigner.php
@@ -42,7 +42,7 @@ class Swift_Signers_OpenDKIMSigner extends Swift_Signers_DKIMSigner
     {
         $header = new Swift_Mime_Headers_OpenDKIMHeader('DKIM-Signature');
         $headerVal = $this->dkimHandler->getSignatureHeader();
-        if (!$headerVal) {
+        if ($headerVal === false || is_int($headerVal)) {
             throw new Swift_SwiftException('OpenDKIM Error: '.$this->dkimHandler->getError());
         }
         $header->setValue($headerVal);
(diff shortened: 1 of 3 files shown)


lib/classes/Swift/Signers/OpenDKIMSigner.php
```diff
@@ -42,7 +42,7 @@ public function addSignature(Swift_Mime_SimpleHeaderSet $headers)
     {
         $header = new Swift_Mime_Headers_OpenDKIMHeader('DKIM-Signature');
         $headerVal = $this->dkimHandler->getSignatureHeader();
-        if (!$headerVal) {
+        if ($headerVal === false || is_int($headerVal)) {
```
c960657: Recent versions return a numeric error code on error.

[source code](https://github.com/xdecock/php-opendkim/commit/7968554f6dcc364d52b9b2d0e43934791038ac96#diff-b515dac4b7866120ce49272ba5e2e5dbR806)

lib/classes/Swift/Signers/OpenDKIMSigner.php
```diff
@@ -53,14 +53,10 @@ public function addSignature(Swift_Mime_SimpleHeaderSet $headers)
...
     {
-        $bodyLen = $this->bodyLen;
-        if (is_bool($bodyLen)) {
-            $bodyLen = -1;
-        }
         $hash = 'rsa-sha1' == $this->hashAlgorithm ? OpenDKIMSign::ALG_RSASHA1 : OpenDKIMSign::ALG_RSASHA256;
         $bodyCanon = 'simple' == $this->bodyCanon ? OpenDKIMSign::CANON_SIMPLE : OpenDKIMSign::CANON_RELAXED;
         $headerCanon = 'simple' == $this->headerCanon ? OpenDKIMSign::CANON_SIMPLE : OpenDKIMSign::CANON_RELAXED;
-        $this->dkimHandler = new OpenDKIMSign($this->privateKey, $this->selector, $this->domainName, $headerCanon, $bodyCanon, $hash, $bodyLen);
+        $this->dkimHandler = new OpenDKIMSign($this->privateKey, $this->selector, $this->domainName, $headerCanon, $bodyCanon, $hash, -1);
```
c960657: -1 means “sign it all”. We want that. `$this->bodyLen` has not been initialised at this point.

lib/classes/Swift/Signers/OpenDKIMSigner.php
```diff
@@ -172,7 +168,7 @@ protected function canonicalizeBody($string)
         if (!$this->peclLoaded) {
             return parent::canonicalizeBody($string);
         }
-        if (false && true === $this->dropFirstLF) {
+        if (true === $this->dropFirstLF) {
```
c960657: We need this. The CRLF being dropped is added in [SimpleMimeEntity::bodyToByteStream()](https://github.com/swiftmailer/swiftmailer/blob/master/lib/classes/Swift/Mime/SimpleMimeEntity.php#L540).

fabpot:
Thank you @c960657.

fabpot:
If the extension is broken, let's deprecate this class in favor of the other one. Can you do a PR for that?

PR #1067, opened 2018-03-15
Deprecate Swift_Signers_OpenDKIMSigner

c960657:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Doc update?   | no
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets |
| License       | MIT

As mentioned in #1066, `Swift_Signers_OpenDKIMSigner` is severely broken. It depends on a [PHP extension](https://github.com/xdecock/php-opendkim) that has no documentation, and is not included in any OS distribution (AFAIK).

Prior to the fixes in #1066, as reported in #640, it did not work at all — at least with not the latest version of the php-opendkim extension.

I didn't find a way to use `relaxed` for the body — this throws an error `OpenDKIM Error: private key load failure` no matter what I do.

This PR deprecates `Swift_Signers_OpenDKIMSigner` in favour of `Swift_Signers_DKIMSigner`.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -6,6 +6,7 @@ Changelog
 
  * added support for non-ASCII email addresses
  * introduced new dependencies: transport.smtphandlers and transport.authhandlers
+ * deprecated Swift_Signers_OpenDKIMSigner; use Swift_Signers_DKIMSigner instead
 
 6.0.2 (2017-09-30)
 ------------------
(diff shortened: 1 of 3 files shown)


fabpot:
Thank you @c960657.

PR #1081, opened 2018-04-14
Correct doc blocks of signer methods

mbabker:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc update?   | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | N/A
| License       | MIT

The short descriptions for the `attachSigner` and `detachSigner` methods are on the wrong methods.  This is now corrected.


fabpot:
Good catch, thanks @mbabker.

Issue #1087, opened 2018-05-17
Use of undefined constant INTL_IDNA_VARIANT_UTS46


iatanasov77:
| Q                   | A
| ------------------- | -----
| Bug report?         | yes
| Feature request?    | no
| RFC?                | no
| How used?           | Laravel
| Swiftmailer version | 6.1.0
| PHP version         | 7.1.17

### Example

I send email by smtp transport with SwiftMailer. 
Mail::send( 'emails.contact', $data,
    function ( $message )
    {
        $message->to( 'noreply@mysite.bg' );
        $message->subject( 'Mail Subject' );
    }
);
Thе code worked before it was updated from 6.0.2 to 6.1.0 version

### Throws an Exception
[2018-05-17 06:06:56] production.ERROR: Use of undefined constant INTL_IDNA_VARIANT_UTS46 - assumed 'INTL_IDNA_VARIANT_UTS46' {"exception":"[object] (ErrorException(code: 0): Use of undefined constant INTL_IDNA_VARIANT_UTS46 - assumed 'INTL_IDNA_VARIANT_UTS46' at /home/lafkabgu/Lafka_2017/vendor/swiftmailer/swiftmailer/lib/classes/Swift/AddressEncoder/IdnAddressEncoder.php:55)
[stacktrace]
#0 /home/lafkabgu/Lafka_2017/vendor/swiftmailer/swiftmailer/lib/classes/Swift/AddressEncoder/IdnAddressEncoder.php(55): Illuminate\\Foundation\\Bootstrap\\HandleExceptions->handleError(8, 'Use of undefine...', '/home/lafkabgu/...', 55, Array)
#1 /home/lafkabgu/Lafka_2017/vendor/swiftmailer/swiftmailer/lib/classes/Swift/AddressEncoder/IdnAddressEncoder.php(43): Swift_AddressEncoder_IdnAddressEncoder->idnToAscii('lafka.bg')
#2 /home/lafkabgu/Lafka_2017/vendor/swiftmailer/swiftmailer/lib/classes/Swift/Transport/EsmtpTransport.php(377): Swift_AddressEncoder_IdnAddressEncoder->encodeString('noreply@lafka.b...')
#3 /home/lafkabgu/Lafka_2017/vendor/swiftmailer/swiftmailer/lib/classes/Swift/Transport/AbstractSmtpTransport.php(478): Swift_Transport_EsmtpTransport->doMailFromCommand('noreply@lafka.b...')
[...]

PR #1088, opened 2018-05-17
Fast fix of issue #1087

iatanasov77:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc update?   | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1087
| License       | MIT

### DIFF
--- a/lib/classes/Swift/AddressEncoder/IdnAddressEncoder.php
+++ b/lib/classes/Swift/AddressEncoder/IdnAddressEncoder.php
@@ -51,9 +51,13 @@ class Swift_AddressEncoder_IdnAddressEncoder implements Swift_AddressEncoder
      */
     protected function idnToAscii(string $string): string
     {
-        if (function_exists('idn_to_ascii') && defined('INTL_IDNA_VARIANT_UTS46')) {
-            return idn_to_ascii($string, 0, INTL_IDNA_VARIANT_UTS46);
-        }
+        if (function_exists('idn_to_ascii')) {
+			if (defined('INTL_IDNA_VARIANT_UTS46')) {
+				return idn_to_ascii($string, IDNA_DEFAULT, INTL_IDNA_VARIANT_UTS46);
+			} else {
+				return idn_to_ascii($string);
+			}
+		}
 
         if (class_exists('TrueBV\Punycode')) {
             $punycode = new \TrueBV\Punycode();
--- a/lib/classes/Swift/AddressEncoder/IdnAddressEncoder.php
+++ b/lib/classes/Swift/AddressEncoder/IdnAddressEncoder.php
@@ -51,7 +51,7 @@ class Swift_AddressEncoder_IdnAddressEncoder implements Swift_AddressEncoder
      */
     protected function idnToAscii(string $string): string
     {
-        if (function_exists('idn_to_ascii')) {
+        if (function_exists('idn_to_ascii') && defined('INTL_IDNA_VARIANT_UTS46')) {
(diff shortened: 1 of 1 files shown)


c960657:
AFAICT the constants `INTL_IDNA_VARIANT_2003` and `INTL_IDNA_VARIANT_UTS46` were introduced in PHP 5.4.0 ([this commit](https://github.com/php/php-src/commit/fa4fe639447e821a0617b80609f18eb251d4bf2b)). However, `INTL_IDNA_VARIANT_UTS46` depends on the `HAVE_46_API` compile-time flag, so I guess this is the problem.

I don't know the difference between the two variants, but in most cases they will give the same result. So I suggest we just fall back to `INTL_IDNA_VARIANT_2003`, if `INTL_IDNA_VARIANT_UTS46` is not defined.

Rotzbua:
`INTL_IDNA_VARIANT_2003` is deprecated since php7.2
https://wiki.php.net/rfc/deprecate-and-remove-intl_idna_variant_2003

c960657:
True, but it is still available in PHP < 8. When `INTL_IDNA_VARIANT_2003 ` is removed in PHP 8, I guess it's safe to assume that `INTL_IDNA_VARIANT_UTS46` is always defined. Otherwise `idn_to_ascii()` wont work at all.

c960657:
Just to clarify, my suggestion is this (assuming that the constant `INTL_IDNA_VARIANT_2003` still exists in PHP 8 but is just not supported by `idn_to_ascii`):

```
return idn_to_ascii($string, 0, defined('INTL_IDNA_VARIANT_UTS46') ? INTL_IDNA_VARIANT_UTS46 : INTL_IDNA_VARIANT_2003);
```

or

```
if (function_exists('idn_to_ascii')) {
    if (defined('INTL_IDNA_VARIANT_UTS46'))
        return idn_to_ascii($string, IDNA_DEFAULT, INTL_IDNA_VARIANT_UTS46);
    } else {
        return idn_to_ascii($string);
    }
}
```

iatanasov77:
I agree with last suggestion of Christian. I test on server with defined INTL_IDNA_VARIANT_UTS46 and on another server where not defined INTL_IDNA_VARIANT_UTS46. It's work.

lib/classes/Swift/AddressEncoder/IdnAddressEncoder.php
```diff
@@ -52,7 +52,11 @@ public function encodeString(string $address): string
     protected function idnToAscii(string $string): string
     {
         if (function_exists('idn_to_ascii')) {
-            return idn_to_ascii($string, 0, INTL_IDNA_VARIANT_UTS46);
+            if (defined('INTL_IDNA_VARIANT_UTS46')) {
+                return idn_to_ascii($string, IDNA_DEFAULT, INTL_IDNA_VARIANT_UTS46);
+            } else {
+                return idn_to_ascii($string);
```
Rotzbua: To clearify maybe add a comment that IDNA2003  is used?

lib/classes/Swift/AddressEncoder/IdnAddressEncoder.php
```diff
@@ -52,7 +52,11 @@ public function encodeString(string $address): string
...
+            }
         }
 
         if (class_exists('TrueBV\Punycode')) {
```
Rotzbua: The polyfil just uses IDNA2003 also comment?

fabpot:

PR #1091, opened 2018-05-25
always return arrays (instead of null) for headers that should return arrays fixes #693

### LINKED ISSUE #693: [5.X] BUG - Return types for getTo(), getCc() and getBcc()
`\Swift_Mime_SimpleMessage` implements the `\Swift_Mime_Message` interface, which states the following functions should return `string[]` arrays only:
- `\Swift_Mime_Message::getTo()`
- `\Swift_Mime_Message::getCc()`
- `\Swift_Mime_Message::getBcc()`

But these functions may return `null` instead of empty arrays. The interface is thus not respected and a bug can occur.

@fabpot @xdecock , I am willing to provide a PR to fix that, but which way round ?
- Shall the interface be updated to indicate these functions can also return `null` ?
- OR Shall the concrete implementations be updated so that they only return `string[]` arrays ?

thewunder:
Q | A
-- | --
Bug fix? | yes
New feature? | no
Doc update? | yes
BC breaks? | yes
Deprecations? | no
Fixed tickets | #693 
License | MIT

### DIFF
--- a/lib/classes/Swift/Mime/SimpleMessage.php
+++ b/lib/classes/Swift/Mime/SimpleMessage.php
@@ -170,7 +170,7 @@ class Swift_Mime_SimpleMessage extends Swift_Mime_MimePart
      */
     public function getSender()
     {
-        return $this->getHeaderFieldModel('Sender');
+        return $this->getHeaderFieldModel('Sender', []);
     }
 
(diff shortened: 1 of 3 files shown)


fabpot:

Issue #1100, opened 2018-07-03
No IDN encoder found (install the intl extension or the true/punycode package


m144:
| Q                   | A
| ------------------- | -----
| Bug report?         | yes
| Feature request?    | no
| RFC?                | no
| How used?           | laravel
| Swiftmailer version |  6.1.0
| PHP version         | 7.0.30

When upgrading swiftmailer version from 6.0.2 to 6.1.0 , I get the following error when trying to send an email:
exception: "Swift_SwiftException"
file:"vendor/swiftmailer/swiftmailer/lib/classes/Swift/AddressEncoder/IdnAddressEncoder.php"
line:64
message: "No IDN encoder found (install the intl extension or the true/punycode package"

This is the first time it happened after using this package for a long time.

I think the correct way to solve this is to make the php extension ext-intl or the package true/punycode a requirement, not only a suggestion, in composer.json

(When installing the true/punycode package the issue was solved)

c960657:
Which email address are you trying to send to? Can you post the full stack trace?

The exception thrown by `Swift_AddressEncoder_IdnAddressEncoder` is a subclass of `Swift_RfcComplianceException`. `Swift_RfcComplianceException` is also thrown if you are trying to send to a malformed email address, so you probably want to catch it in any case, unless you have validated the addresses in advance.

m144:
All validations are done before sending, and I have been sending to the same email address I sent before with success. 
I know this is meant to solve a problem with utf-8 characters in the email address, but I have none in my addresses.
Here is the stack trace:
```
{
    "message": "No IDN encoder found (install the intl extension or the true/punycode package",
    "exception": "Swift_SwiftException",
    "file": "vendor/swiftmailer/swiftmailer/lib/classes/Swift/AddressEncoder/IdnAddressEncoder.php",
    "line": 64,
    "trace": [
        {
            "file": "vendor/swiftmailer/swiftmailer/lib/classes/Swift/AddressEncoder/IdnAddressEncoder.php",
            "line": 43,
            "function": "idnToAscii",
            "class": "Swift_AddressEncoder_IdnAddressEncoder",
            "type": "->"
        },
        {
            "file": "vendor/swiftmailer/swiftmailer/lib/classes/Swift/Mime/Headers/IdentificationHeader.php",
            "line": 164,
            "function": "encodeString",
            "class": "Swift_AddressEncoder_IdnAddressEncoder",
            "type": "->"
        },
        {
[...]

PR #1102, opened 2018-07-03
To not require IDN encoder for ASCII-only address

c960657:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc update?   | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1100
| License       | MIT

Do not require an IDN encoder when passed an ASCII-only address.

Throw `Swift_AddressEncoderException` if passed a non-ASCII address and no IDN encoder is present. This is consistent with the existing functionality that throws this exception when passed an address that is not supported by the current encoder.


taylorotwell:
Nice.

fabpot:
Thank you @c960657.

PR #1105, opened 2018-07-07
Fix return type in MessageLogger

kubk:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc update?   | yes
| BC breaks?    | no
| Deprecations? | no

Method `Swift_Events_SendEvent::getMessage()` returns `Swift_Mime_SimpleMessage`
https://github.com/swiftmailer/swiftmailer/blob/master/lib/classes/Swift/Events/SendEvent.php#L77


fabpot:
Good catch, thanks @kubk.

Issue #1106, opened 2018-07-09
Changed behaviour using SMTP Transport


svandervlugt:
| Q                   | A
| ------------------- | -----
| Bug report?         | yes
| Feature request?    | no
| RFC?                | no
| How used?           | Symfony
| Swiftmailer version | 6.1
| PHP version         | 7.1

### Observed behaviour
When using smtp as transport layer, sending a message to a rejected e-mail-address throws an Swift_TransportException.

### Expected behaviour
Before version 6.1 \Swift_Transport->send() would return 0 when a mail-server returned recipient address rejected: Domain not found.

The cause seems to be the default enabling of pipelining when a mail-server has the capability for it.

PR #1107, opened 2018-07-09
Handle recipient errors when pipelining

c960657:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc update?   | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1106
| License       | MIT

When using SMTP PIPELINING, unexpected replies may throw exceptions in the function that cause the pipeline to be processed (usually `doDataCommand()`) rather than the command that sends the offending command (e.g. `doRcptToCommand()`).

This is a problem in `doMailTransaction()` that wants to catch exceptions caused by invalid recipients rejected in `doRcptToCommand()`.

In fact, this issue was a BC break introduced in #1065. Would the appropriate fix be to roll back this feature for the 6.x branch and postpone it to 7.x?

I would like to give this patch some though. I'll post an update in a few days.

### DIFF
--- a/lib/classes/Swift/Transport/AbstractSmtpTransport.php
+++ b/lib/classes/Swift/Transport/AbstractSmtpTransport.php
@@ -313,15 +313,18 @@ abstract class Swift_Transport_AbstractSmtpTransport implements Swift_Transport
      *
      * If no response codes are given, the response will not be validated.
      * If codes are given, an exception will be thrown on an invalid response.
+     * If the command is RCPT TO, and the pipeline is non-empty, no exception
+     * will be thrown; instead the failing address is added to $failures.
      *
      * @param string   $command
      * @param int[]    $codes
      * @param string[] $failures An array of failures by-reference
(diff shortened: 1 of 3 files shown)


fabpot:
Thanks for working on this. I would prefer to find a fix and do a patch release instead of reverting the feature. If you are not confident enough, reverting is also an option of course. Let me know.

c960657:
Ok, I have looked it through again, and I think it's ready to fly.

fabpot:
Thank you @c960657.

spronkey:
@c960657  Not sure whether to open a new ticket or comment here, but I'm not sure this issue is 100% fixed. If there are _no_ valid recipients, sending a DATA command can result in the server responding with a 554 no valid recipients error instead of the expected 354.

SwiftMailer will then throw that TransportException exception instead. So it's still throwing a transport exception instead of just returning 0.

PR #1112, opened 2018-07-30
wxu: Fix release version is not compatible with low version PHP

Isites:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc update?   | no
| BC breaks?    | no
| Deprecations? | no

https://github.com/Isites/swiftmailer/blob/master/lib/classes/Swift/Transport/StreamBuffer.php#L95

### DIFF
--- a/lib/classes/Swift/Transport/StreamBuffer.php
+++ b/lib/classes/Swift/Transport/StreamBuffer.php
@@ -95,7 +95,9 @@ class Swift_Transport_StreamBuffer extends Swift_ByteStream_AbstractFilterableIn
         // To support modern tls we allow explicit tls1.0, tls1.1, tls1.2
         // Ssl3 and older are not allowed because they are vulnerable
         // @TODO make tls arguments configurable
-        return stream_socket_enable_crypto($this->_stream, true, STREAM_CRYPTO_METHOD_TLSv1_0_CLIENT | STREAM_CRYPTO_METHOD_TLSv1_1_CLIENT | STREAM_CRYPTO_METHOD_TLSv1_2_CLIENT);
+	// To Support more php versions
+	$supportTls = PHP_VERSION > '5.6' ? STREAM_CRYPTO_METHOD_TLSv1_0_CLIENT | STREAM_CRYPTO_METHOD_TLSv1_1_CLIENT | STREAM_CRYPTO_METHOD_TLSv1_2_CLIENT : STREAM_CRYPTO_METHOD_TLS_CLIENT;
+        return stream_socket_enable_crypto($this->stream, true, $supportTls);
     }
 
     /**


lib/classes/Swift/Transport/StreamBuffer.php
```diff
@@ -91,7 +91,9 @@ public function startTLS()
         // To support modern tls we allow explicit tls1.0, tls1.1, tls1.2
         // Ssl3 and older are not allowed because they are vulnerable
         // @TODO make tls arguments configurable
-        return stream_socket_enable_crypto($this->stream, true, STREAM_CRYPTO_METHOD_TLSv1_0_CLIENT | STREAM_CRYPTO_METHOD_TLSv1_1_CLIENT | STREAM_CRYPTO_METHOD_TLSv1_2_CLIENT);
+	// To Support more php versions
+	$supportTls = PHP_VERSION > '5.6' ? STREAM_CRYPTO_METHOD_TLSv1_0_CLIENT | STREAM_CRYPTO_METHOD_TLSv1_1_CLIENT | STREAM_CRYPTO_METHOD_TLSv1_2_CLIENT : STREAM_CRYPTO_METHOD_TLS_CLIENT;
```
fabpot: I don't think it's going to work. I would do something like:

```
if (!defined('STREAM_CRYPTO_METHOD_TLSv1_0_CLIENT')) {
    define('STREAM_CRYPTO_METHOD_TLSv1_0_CLIENT', 0);
}

```

Isites: I have tried another way. 

```php
defined('STREAM_CRYPTO_METHOD_TLSv1_0_CLIENT') || define('STREAM_CRYPTO_METHOD_TLSv1_0_CLIENT', 3);
defined('STREAM_CRYPTO_METHOD_TLSv1_1_CLIENT') || define('STREAM_CRYPTO_METHOD_TLSv1_1_CLIENT', 0);
defined('STREAM_CRYPTO_METHOD_TLSv1_2_CLIENT') || define('STREAM_CRYPTO_METHOD_TLSv1_2_CLIENT', 0);
```

this will be successful

fabpot:
Thank you, I'm going to take over and make a release today.

moisis-dm:
There is a bug on the commit you made today
Undefined property $stream

This should change
`return stream_socket_enable_crypto($this->stream, true, $cryptoType);`
to
`return stream_socket_enable_crypto($this->_stream, true, $cryptoType);`

fabpot:
I knew it was a bad idea to backport that. I've already spent too much time on this. New version has just been released (thanks @moisis-dm for the report).

lib/classes/Swift/Transport/StreamBuffer.php
```diff
@@ -95,7 +95,9 @@ public function startTLS()
         // To support modern tls we allow explicit tls1.0, tls1.1, tls1.2
         // Ssl3 and older are not allowed because they are vulnerable
         // @TODO make tls arguments configurable
-        return stream_socket_enable_crypto($this->_stream, true, STREAM_CRYPTO_METHOD_TLSv1_0_CLIENT | STREAM_CRYPTO_METHOD_TLSv1_1_CLIENT | STREAM_CRYPTO_METHOD_TLSv1_2_CLIENT);
+	// To Support more php versions
+	$supportTls = PHP_VERSION > '5.6' ? STREAM_CRYPTO_METHOD_TLSv1_0_CLIENT | STREAM_CRYPTO_METHOD_TLSv1_1_CLIENT | STREAM_CRYPTO_METHOD_TLSv1_2_CLIENT : STREAM_CRYPTO_METHOD_TLS_CLIENT;
```
xabbuh: Shouldn't you rather use `PHP_VERSION_ID` instead?

Isites: I am careless, thank you for pointing out my mistake. @xabbuh 
Thank you for maintaining this project. @fabpot

xabbuh: Ah, I missed that this was already fixed by d9aed422e20f2191d5d4324fce102b3db9b5de6b.

PR #1128, opened 2018-09-17
re-queue message when an exception occurred

armetiz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc update?   | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

re-queue message when all retries failed and when an other type of exception occurred.


fabpot:

Issue #737, opened 2016-03-16
offering openpgp plugin for swiftmailer 


postme:
[OpenPGPSigner.zip](https://github.com/swiftmailer/swiftmailer/files/181227/OpenPGPSigner.zip)

Hi there,
for one of my projects I needed an OpenPGP signing/encrypting solution for Swiftmailer. I have the code available and am happy to donate it to Swiftmailer. Its a standard Signer, implementing  Swift_Signers_BodySigner. Are you interested in receiving the code and making it part of Swiftmailer?
Kind regards
Meint

rubiodamian:
+1

fritzmg:
This signer should definitely be integrated. Any updates on this? No comments from Swiftmailer devs?

pavarnos:
See also https://github.com/Mailgarant/switfmailer-openpgp

PR #1136, opened 2018-10-05
Add OpenPGP message signer

digilist:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc update?   | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #737 
| License       | MIT

I added a OpenPGP message signer that is able to sign and encrypt messages. My code is based on the one that @postme provided in #737, but I cleaned it up and added tests.

Enigmail/Thunderbird was able to decrypt and verify the generated messages. However, I am not totally sure if everything is correct.

One thing I am especially not sure about is the micalg header parameter. The parameter must define the used hashing algorithm, but PHP does not allow to define the algorithm. Currently the code names SHA256 as micalg (which I just adopted from @postme's code), so I am not sure about this.

In my tests with Enigmail the micalg value didn't matter at all. I could remove it and Enigmail still said my signature was good. However, according to [RFC 3156](https://tools.ietf.org/html/rfc3156) the micalg parameter is required, so I do not think it should stay this way.

> The "micalg" parameter for the "application/pgp-signature" protocol MUST contain exactly one hash-symbol of the format "pgp-<hash-identifier>", where <hash-identifier> identifies the Message Integrity Check (MIC) algorithm used to generate the signature.

Update: I do not really know, why the unit tests with travis are failing because of missing keys. On my local computer they are working. Probably a GnuPG version mismatch?


Rotzbua:
Interesting pr. But test fails. Seems incomplete.

> Swift_IoException: Unable to find an active key to encrypt for test@example.org, try importing keys first

digilist:
Yes, that's true and unfortunately I do not know why. Locally those tests are running and I think it might be a different PGP version, but I do not know how to confirm that. I was hoping, someone else got an idea.

I am using this signer in another project in production and it's working. It's unit tested there as well and locally all tests are running and the signing works, but we experienced the same failing test behavior behavior with Scrutinizer.

lib/classes/Swift/Signers/OpenPGPSigner.php
```diff
@@ -0,0 +1,404 @@
...
+        );
+    }
+
+    /**
+     * Checks whether the given key is still valid.
+     *
+     * @param array  $key
+     * @param string $purpose
+     *
+     * @return bool|mixed
```
Rotzbua: should be only bool

lib/classes/Swift/Signers/OpenPGPSigner.php
```diff
@@ -0,0 +1,404 @@
...
+            return $this;
+        }
+
+        $plaintextBody = $this->extractPlainTextBody($message);
+
+        if ($this->sign) {
+            $plaintextBody = $this->pgpSignMessage($message, $plaintextBody);
+
+            // To use the signed body for encryption, we need to prepend the Content-Type header into the plaintext
+            $plaintextBody = sprintf("%s\r\n%s", $message->getHeaders()->get('Content-Type')->toString(), $plaintextBody);
```
Rotzbua: single quote

lib/classes/Swift/Signers/OpenPGPSigner.php
```diff
@@ -0,0 +1,404 @@
...
+        if (!class_exists(gnupg::class)) {
+            throw new RuntimeException('Please install gnupg extension to use this signer.');
+        }
+
+        if ($gnupgHome !== null) {
+            if (!file_exists($gnupgHome)) {
+                throw new Swift_SwiftException('GnuPG home path does not exist');
+            }
+
+            putenv("GNUPGHOME=" . escapeshellcmd($gnupgHome));
```
Rotzbua: single quote and success should be checked

tests/unit/Swift/Signers/OpenPGPSignerTest.php
```diff
@@ -0,0 +1,269 @@
...
+    /*private*/ const BEGIN_PGP_SIGNATURE = '-----BEGIN PGP SIGNATURE-----';
+
+    /**
+     * @var gnupg
+     */
+    private $gnupg;
+
+    public function setUp()
+    {
+        putenv("GNUPGHOME=" . escapeshellcmd(self::GNUPG_HOME));
```
Rotzbua: check success

Rotzbua (changes requested):
5 minute review, maybe env vars are not set correctly and the keys are not found

fabpot:

PR #1139, opened 2018-10-29
Mute errors when writing to the stream

willemstuursma:
This allows cooperating with other error handlers that convert errors into ErrorExceptions. Success in writing to the stream is checked at the next line.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc update?   | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

Ran into an issue where writing into the stream failed. `fwrite` triggered an error, which was then converted to an `ErrorException` by one of the error handlers in my application. Hence, sending the email failed.

Since the return value of `fwrite` is already checked, there is no need to also trigger an error so we can use the silence operator.


fabpot:

PR #1148, opened 2018-11-23
Update transport_deps.php

wandesoncunha:
when we use php-cli the variable SERVER_NAME does not exist. Need to use gethostname() to fetch the client host.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc update?   | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #... 
| License       | MIT

when we use php-cli the variable SERVER_NAME does not exist. Need to use gethostname() to fetch the client host.


fabpot:

PR #1154, opened 2018-12-09
Update travis for version 5

gmponos:


Rotzbua:
version 5 is out of support and will not be changed to work under newer php versions.

gmponos:
> Indeed, version is not maintained anymore, let's close.

thank you.. I asked also on slack but I didn't have an answer... 😄

PR #1158, opened 2018-12-31
Don't throw on DATA when all recipients in pipeline fail

c960657:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc update?   | no
| BC breaks?    | yes (small)
| Deprecations? | no
| Fixed tickets |
| License       | MIT

As reported by @spronkey in https://github.com/swiftmailer/swiftmailer/pull/1107#issuecomment-430460153 there is a bug in SMTP pipelining:
> @c960657 Not sure whether to open a new ticket or comment here, but I'm not sure this issue is 100% fixed. If there are no valid recipients, sending a DATA command can result in the server responding with a 554 no valid recipients error instead of the expected 354.
>
>SwiftMailer will then throw that TransportException exception instead. So it's still throwing a transport exception instead of just returning 0.


fabpot:

PR #1188, opened 2019-03-31
fix(SimpleHeaderSet): invalid return type for `get()`

jacekkarczmarczyk:
phpdoc comment says the `get` method returns NULL if the header is not present, but it's not reflected in the `@return`

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc update?   | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | n/a
| License       | MIT


jacekkarczmarczyk:
Duplicate of https://github.com/swiftmailer/swiftmailer/pull/1260

PR #1210, opened 2019-07-23
Fix RfcComplianceException handling on messages without To field

Tamrael:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc update?   | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1209 
| License       | MIT

See issue #1209 for description of the bug.
This patch does the following:

allow for messages without To field
include Cc- and Bcc-Fields into failedRecipients array on RfcComplianceException


fabpot:

PR #1227, opened 2019-10-14
bugfix and code refactor: $response var might be initialized fixed; nested condition is converted to guard clause in to improve code readability

lebedevsergey:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes/no
| Doc update?   | yes/no
| BC breaks?    | yes/no
| Deprecations? | yes/no
| Fixed tickets | #... 
| License       | MIT

$response var might be initialized - added its initialization to null
also nested condition is converted to guard clause in order to improve code readability


fabpot:
Thank you @lebedevsergey.

Issue #311, opened 2013-04-04
Some issues with the file spool when messages fail


bicpi:
Recently I encountered some difficulties with the file spool feature (in conjunction with SwiftmailerBundle's `SendEmailCommand` command). This was also my topic on the KnpUniversity Q&A day and my question was analysed in depth at https://knpuniversity.com/screencast/question-answer-day/swiftmailer-spooling. The solution was a workaround and I'm currently helping myself with a similar approach.

But nonetheless I think the `Swift_FileSpool` may need some improvements:
1. What is the reason for `Swift_FileSpool`'s attributes `$_path` and `$_retryLimit` being declared as `private`? To implement my workaround in a customized child class I was struggling with overriding some methods (`recover()` and `flushQueue()`) because I have no  access to these properties. So I had to copy all the coode down to my child class. I could not use an independent class because SwiftmailerBundle's `SendEmailCommand` command requires an `Swift_FileSpool` instance.
2. `Swift_FileSpool::flushQueue()`: I'm using Amazons SES service via SMTP. When sending out thousands of mails every day it is likely to sometimes receive a permanent  SMTP error for some reasons (e.g. _recepient blacklisted_ or _sender not verified_). This causes `$transport->send()` to throw an `Swift_TransportException` which is not catched and kills the remaining sendouts. There are four problems with this:
   - You won't get informed about the sent mails before the error (`$count` variable)
   - You won't get informed about what mail has failed
   - No more mails are sent after the error, but it seems that we could recover from an error for a particular mail - all the mails afterwards could be ok and would not need to wait for  the next queue flush.
   - Supposed to flush the queue every minute via cron and supposed to use the default interval for recovery (15 minutes): As soon as more than 15 messages with permanent errors have accumulated in the spool directory - no email at all will be sent any more. This is because the first failed message recovers afters 15 minutes and fails again with the sending process completely aborted. One minute later it's the same with the second message and so on.  

In my workaround I've created a customized child class of `Swift_FileSpool`. I'm catching  `Swift_TransportException`s for the `$transport->send()` call in `flushQueue()` to continue with the "good"  messages in the same process - currently not beeing able to log that. To get rid of permanently failing messages I've added a second parameter `$recoverLimit=3` to the `recover()` method. It is used to indicate how many attempts for recovery should be made. All spooled message files are removed after `$recoverLimit*$timeout` seconds.

[...]

PR #1230, opened 2019-10-30
Added message expiration feature.

cyberemissary:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Doc update?   | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #311
| License       | MIT

This PR adds a message expiration feature. Basically the idea is to eventually stop trying to resend the same message over and over again if it keep failing (which eventually results in the issue described in #311). 

File spool will now add an *attempt* number to the message filename and use that to track how many delivery attempts were done on that particular file. 

When the maximum retry attempts have been reached the failed message will be moved to `expired` directory inside the spool directory, so administrators can potentially review those messages if needed.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,7 @@
 Changelog
 =========
+6.3.0 (2019-XX-XX)
+ * Added message expiration feature @cyberemissary
 
 6.2.2 (2019-XX-XX)
 ------------------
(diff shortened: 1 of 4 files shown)


klandaika:
It would be nice if this feature got merged, We currently are forced to implement workarounds on all of our projects to avoid overflow of bad emails.

cyberemissary:
Good day @fabpot and @xdecock 
Do you see any issues with this feature? I would really like to see this through.

Thanks

fabpot:

PR #1243, opened 2019-12-20
[WIP] [6.x] Test on PHP 7.4

GrahamCampbell:


fabpot:
@GrahamCampbell Will you fix the tests for 7.4 and do you want me to take over?

GrahamCampbell:
Just seen the build output now. I think it's likely at least the first 22 of these are actually due to PHPUnit itself. I don't have time to look into how hard also supporting PHPUnit 7 (7 is enough I think to fix the PHP 7.4 issues) right now, but I can do it later today if you don't beat me to it.

fabpot:
Taking over #1252

PR #1251, opened 2020-02-07
Sleep fixes

Manriel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc update?   | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 
| License       | MIT

Sleep can be interrupted by pcntl signal receiving and returns the number of seconds left to sleep.

### DIFF
--- a/lib/classes/Swift/Plugins/AntiFloodPlugin.php
+++ b/lib/classes/Swift/Plugins/AntiFloodPlugin.php
@@ -131,7 +131,9 @@ class Swift_Plugins_AntiFloodPlugin implements Swift_Events_SendListener, Swift_
         if (isset($this->sleeper)) {
             $this->sleeper->sleep($seconds);
         } else {
-            sleep($seconds);
+            do {
+                $seconds = sleep($seconds);
+            } while ($seconds > 0);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


lib/classes/Swift/Plugins/AntiFloodPlugin.php
```diff
@@ -131,7 +131,9 @@ public function sleep($seconds)
         if (isset($this->sleeper)) {
             $this->sleeper->sleep($seconds);
         } else {
-            sleep($seconds);
+            do {
+                $seconds = sleep($seconds);
```
fabpot: The PHP docs mention that on Windows, it always returns 192.
Do you have any use cases in mind that makes this change required?

Manriel: In case the SwiftMailer instanced as a server worker that handle the sending queue and the sending is being throttled, the SIGUSR1 call interrupt the sleeping, that perform the excess with sending.

As far as I know, Windows don't use POSIX signals, so we can implement some condition handling that case. Anyhow, the algo of default sleep becaming complex. Maybe it make sense to implement a Default Sleeper instead.

fabpot:

PR #1255, opened 2020-02-17
fix indentation

DurandSacha:


fabpot:
For everything related to coding standard, we are relying on PHP-CS-Fixer to automate the task. So, if something is not implemented by PHP-CS-Fixer, we don't update the code (and again, we would have to do it everywhere instead of just one place).

PR #1260, opened 2020-03-05
fix: wrong phpdoc for Swift_Mime_SimpleHeaderSet::get() method

jacekkarczmarczyk:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc update?   | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #... 
| License       | MIT

As per https://github.com/swiftmailer/swiftmailer/blob/master/lib/classes/Swift/Mime/SimpleHeaderSet.php#L190 `Swift_Mime_SimpleHeaderSet::get()` may return `Swift_Mime_Header|null`, but phpdoc states that it always returns `Swift_Mime_Header` which is not true and which breaks code validation tools like phpstan


fabpot:
Thank you @jacekkarczmarczyk.

Issue #1265, opened 2020-03-12
Failing DATA command in pipelining should send a RSET


garry08:
| Q                   | A
| ------------------- | -----
| Bug report?         | yes
| Feature request?    | no
| RFC?                | yes
| How used?           | Standalone
| Swiftmailer version | latest
| PHP version         | 7.4.2

### Context

I started my investigation due to the following error message, which led me to [this issue (#1143)](https://github.com/swiftmailer/swiftmailer/issues/1143) (closed without solution lacking sufficient information).

    postfix/smtpd[68279]: improper command pipelining after DATA from example.com[xxx.xxx.xxx.xxx]

### Observed behaviour

Using pipelining, when one sends a `MAIL` command without proper recipients, the `DATA` command will receive an error response which is not handled properly causing nested `MAIL` command problems afterwards.

### Expected behaviour

In such scenario, a `RSET` command should be sent after the failing `DATA` command so a new `MAIL` command can be issued.

Here is an excerpt of the [RFC (2920)](https://tools.ietf.org/html/rfc2920#section-3.1) about this

>Client SMTP implementations that employ pipelining MUST check ALL
   statuses associated with each command in a group. For example, if
   none of the RCPT TO recipient addresses were accepted the client must
   then check the response to the DATA command -- the client cannot
   assume that the DATA command will be rejected just because none of
   the RCPT TO commands worked.  If the DATA command was properly
   rejected the client SMTP can just issue RSET

### Example

Say one writes a script sending two emails one after the other, this might lead to the following SMTP dialogue.

```
<< 220 foo.com SMTP service ready
>> EHLO bar.com
<< 250-foo.com greets bar.com
<< 250-PIPELINING
>> MAIL FROM:<noreply@bar.com>
>> RCPT TO:<non-existent@foo.com>
>> DATA
<< 250 OK
<< 550 No such user here
<< 554 no valid recipients
>> MAIL FROM:<noreply@bar.com>
>> RCPT TO:<existent@foo.com>
>> DATA
<< 503 Bad sequence of commands
>> QUIT
<< 221 goodbye
```

As you can see both email will fail but only the first is a legitimate error. In other words in a single send the second would have worked.

And this can be fixed using the `RSET` command, which would produce

```
<< 220 foo.com SMTP service ready
>> EHLO bar.com
<< 250-foo.com greets bar.com
<< 250-PIPELINING
>> MAIL FROM:<noreply@bar.com>
[...]

PR #1266, opened 2020-03-12
Send RSET command when DATA commands fails.

garry08:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc update?   | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | https://github.com/swiftmailer/swiftmailer/issues/1265
| License       | MIT

When using SMTP PIPELINING, unexpected replies may throw exceptions only when the pipeline is sent (in `doDataCommand()`). In that case the `MAIL` command remains open and might cause `503 Bad sequence of commands` error when the next `MAIL` command is issued.

Applied solution here is to send a `RSET` command in such case.

This is my first commit in this repository, feedback are more than welcome 🙂


fabpot:

PR #1270, opened 2020-04-02
Add flock lock on message during queue flushing

Dcoutelle:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc update?   | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

If you run multiple process of queue flushing ( `swiftmailer:email:send` in swift mailer bundle for exemple), there is a chance that you send the same message 2 time (ratio of 0.1% with 5 process in parallel).
With php rename function, if process2 try to rename a file while process1 doing it, process2 will duplicate file and process1 will wait until process2 finished to delete the original file.

This PR add a flock lock on message before renaming from `xxx.message` to `xxx.message.sending` to avoid this.

### DIFF
--- a/lib/classes/Swift/FileSpool.php
+++ b/lib/classes/Swift/FileSpool.php
@@ -162,17 +162,24 @@ class Swift_FileSpool extends Swift_ConfigurableSpool
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


fabpot:

Issue #1272, opened 2020-04-09
addTo and setTo (and maybe others methods) can throw exception but there is nothing in the PhpDoc.


VincentLanglet:
| Q                   | A
| ------------------- | -----
| Bug report?         | no
| Feature request?    | no
| RFC?                | yes
| How used?           | Symfony
| Swiftmailer version | master
| PHP version         | 7.4

### Observed behaviour
I recently use 
```
$swiftMailer->setTo('email..foo@gmail.com')
```
And got an exception `Swift_RfcComplianceException`.

Since this issue: https://github.com/swiftmailer/swiftmailer/issues/304, there is documentation about the possible exception. But the phpDoc was not updated.

The phpDoc is useful to detect that you don't catch a possible exception.

`setFieldBodyModel` mention the  exception `Swift_RfcComplianceException`.
But `createMailboxHeader` call `setFieldBodyModel` without to mention the exception,
Then `addMailboxHeader` call `createMailboxHeader` without to mention the exception,
Then `setTo` call `addMailboxHeader` without to mention the exception.

### Expected behaviour
I expected the phpDoc to mention the exception. I can make the PR.

Issue #304, opened 2013-03-19
Documentation improvement suggestion regarding setTo(), Swift_RfcComplianceException and Swift_Validate


Haprog:
I think it would be nice if the documentation mentioned that for example `setTo()` will throw a `Swift_RfcComplianceException` if given an invalid email address. The documentation should mention that you can either validate your email addresses using `Swift_Validate::email($email)` prior to using `setTo()` or that you can wrap the `setTo()` call in a try-catch block to catch the exception in case the emails you give it may contain invalid addresses.

This is especially important when your email addresses come from a source that may contain invalid addresses and you're sending a large batch of email in a loop. In that case one invalid email address will stop your whole email loop if you don't catch the exception early enough or validate your addresses before giving them to `setTo()`.

I spent a lot of time figuring this out since I was using `$mailer->send($message, $failures)` in a loop and thought that `$failures` would catch invalid emails until I realized that it was `setTo()` that was throwing the exception even before reaching the `send()` call.

PR #1275, opened 2020-04-16
Add missing Swift_RfcComplianceException throw tag

VincentLanglet:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Doc update?   | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1272
| License       | MIT

Couple of function are possibly throwing `Swift_RfcComplianceException` ; it was documented, but the phpDoc wasn't up to date. Updating the phpDoc would be useful for static analysis because they're looking for missing `try/catch`.

### DIFF
--- a/lib/classes/Swift/Mime/Headers/MailboxHeader.php
+++ b/lib/classes/Swift/Mime/Headers/MailboxHeader.php
@@ -255,6 +255,8 @@ class Swift_Mime_Headers_MailboxHeader extends Swift_Mime_Headers_AbstractHeader
      * @param string[] $mailboxes
      *
      * @return string[]
+     *
+     * @throws Swift_RfcComplianceException
      */
     protected function normalizeMailboxes(array $mailboxes)
     {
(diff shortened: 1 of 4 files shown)


VincentLanglet:
Hi @fabpot, would you have some time to review my PR ? :)

fabpot:
Not a big fan. Why document this specific exception and not all of them everywhere? Also, we tend (at least in Symfony) to only document (optionnally) exceptions when they are actually thrown in the method, not when a method call can thrown it. So, I'm going to close as I don't want to go down this path and have more PRs adding more PHPDocs. Thank you for proposing.

VincentLanglet:
> Why document this specific exception and not all of them everywhere? 

This was related to https://github.com/swiftmailer/swiftmailer/issues/304.
A code should be documented by himself. Instead of writing the possible returned exception in a specific documentation page, it should be in the phpDoc.

The issue was that I recently use

```
$swiftMailer->setTo('email..foo@gmail.com')
```
And got an exception Swift_RfcComplianceException in production, causing a 500 and all the bad consequences for the user.

The developer won't be aware to use a try/catch if it's not correctly documented.

> Also, we tend (at least in Symfony) to only document (optionnally) exceptions when they are actually thrown in the method, not when a method call can thrown it.

Too bad you're going this way, that doesn't help to correctly catch exceptions. This certainly is easier for you but I now feel like I should write every symfony method in a try/catch because I can trust the phpDoc. 😕

PR #1276, opened 2020-04-24
FileByteStream compatibility with PHP 7.4

fourhundredfour:
| Q                  | A
| ------------- | ---
| Bug fix?         | yes
| New feature?  | no
| Doc update?   | no
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets | #
| License       | MIT

Since the deprecation of returning an empty string in the `fread` function when an error occurs, checking for `false` is required for compatibility with PHP7.4.

See: https://www.php.net/manual/en/migration74.incompatible.php#migration74.incompatible.core.fread-fwrite

### DIFF
--- a/lib/classes/Swift/ByteStream/FileByteStream.php
+++ b/lib/classes/Swift/ByteStream/FileByteStream.php
@@ -81,7 +81,7 @@ class Swift_ByteStream_FileByteStream extends Swift_ByteStream_AbstractFilterabl
 
             // If we read one byte after reaching the end of the file
             // feof() will return false and an empty string is returned
-            if ('' === $bytes && feof($fp)) {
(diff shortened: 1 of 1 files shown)


fabpot:
Shouldn't we also use `@fread()` now instead of `fread()` as the migration doc mention that it now raises a notice?

fourhundredfour:
Since Swiftmailer is a library, I think that the error message should not be suppressed so that the user, which uses swiftmailer in his project, knows that there are for example insufficient permissions to that stream.

fourhundredfour:
Any updates? @fabpot

fabpot:
Thank you @fourhundredfour.

nicolas-grekas (approved):

PR #1283, opened 2020-06-01
Update Mailer.php

abdounikarim:
add some php documentation


fabpot:
We don't add phpdocs when it does not add more information than what is already in the signature.

PR #1312, opened 2020-12-28
add return type to setUp and tearDown methods

abdounikarim:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Doc update?   | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #... 
| License       | MIT

Hello,

I add some return types for tests to launch and run.


fabpot:
We cannot do that change to support older PHP versions. Run the tests with `./vendor/bin/simple-phpunit` instead.

PR #1316, opened 2021-01-06
Allow egulias/email-validator 3

VincentLanglet:
See https://github.com/egulias/EmailValidator/releases/tag/3.0.0 for changelog

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
     ],
     "require": {
         "php": ">=7.0.0",
-        "egulias/email-validator": "^2.0",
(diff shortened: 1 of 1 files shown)


fabpot:
There is a BC break (see https://github.com/symfony/symfony/pull/39685).

VincentLanglet:
> There is a BC break (see [symfony/symfony#39685](https://github.com/symfony/symfony/pull/39685)).

Indeed.
If I understand correctly the discussion, you're waiting for some changes in egulias/email-validator to allow it, isn't it ?

PR #1317, opened 2021-01-12
Don't trust properties at destruct time

jderusse:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Doc update?   | no
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | -
| License       | MIT

Prevent destructors with side-effects from being unserialized

### DIFF
--- a/lib/classes/Swift/ByteStream/TemporaryFileByteStream.php
+++ b/lib/classes/Swift/ByteStream/TemporaryFileByteStream.php
@@ -39,4 +39,14 @@ class Swift_ByteStream_TemporaryFileByteStream extends Swift_ByteStream_FileByte
             @unlink($this->getPath());
         }
     }
+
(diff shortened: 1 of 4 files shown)


fabpot:
Thank you @jderusse.

jderusse:
Thank you @nicolas-grekas for the patch.

pperzyna:
@jderusse @fabpot What is wrong with deserialization? It is a common practice and even more important for emails that are in a queue!

jderusse:
this is can be abused by attacker. See https://owasp.org/www-community/vulnerabilities/PHP_Object_Injection

jderusse:
> more important for emails that are in a queue!

Messages are serializable. Transport and Temporary streams are not (and, they never should have been serialized)

Issue #545, opened 2014-10-21
Misleading SSL/TLS option


dol:
The security argument of Swift_SmtpTransport is misleading with the naming.

```
'ssl' => SMTP over SSL/TLS (recommended port: 465)
'tls' => STARTTLS (SSL/TLS) over TCP (recommended port: 587)
```

`'ssl'` can also use the [TLS protocol](https://www.fastmail.fm/help/technical/ssltlsstarttls.html). Since SSLv3 are widely disabled, only the TLS protocol is used. It would be clearer, if the options are called differently.
This is a BC break, if the security options are called differently.

An possible solution to this problem would be to an autodetection of the SSL or STARTTLS. If the argument 'security' is set to 'true', it first tries to establish an SSL connection, if it's not working, try reestablish an 'TCP' connection and look for the Capability of 'STARTTLS'.

RealRancor:
I agree that this should be clarified. Some notes about the difference between the ssl(SSL/TLS) and the tls(STARTTLS) could be also mentioned at the documentation:

http://swiftmailer.org/docs/sending.html#encrypted-smtp

This has also lead to some confusion in https://github.com/owncloud/documentation/issues/1778

Sebi94nbg:
Thanks guys!

Laravel 5.4 uses this PHP libary too and had the problem, that it couldn't send emails due of a SSLv3 handshake failure:
```
stream_socket_enable_crypto(): SSL operation failed with code 1. OpenSSL Error messages:
error:14094410:SSL routines:ssl3_read_bytes:sslv3 alert handshake failure
```

After changing the mentioned settings, it worked fine:
```
MAIL_PORT=465 (before '25')
MAIL_ENCRYPTION=ssl (before 'tls')
```

Rotzbua:
@Sebi94nbg ssl3 is broken for years now... better upgrade your mailserver.. this issue is about the naming of the technique

Sebi94nbg:
@Rotzbua the mail server is up2date, not vulnerable and does only allow recommended cipher suites and protocols. That's the reason, why I got this issue.

Rotzbua:
@Sebi94nbg I think that a server environment issue. It seems that php and mail server do not support the equal ciphers. Which cipher does your server (php,curl,openssl) and your mailserver support?

Maybe it is this bug, read first comment:
https://secure.php.net/manual/de/function.stream-socket-enable-crypto.php

Sebi94nbg:
[...]

PR #1330, opened 2021-01-27
[doc] Clarify encryption options 'ssl' vs. 'tls'

### LINKED ISSUE #1329: [doc / question] Meaning of $encryption = "ssl" vs. "tls"
| Q                   | A
| ------------------- | -----
| Bug report?         | no
| Feature request?    | yes (extending documentation)
| RFC?                | no
| How used?           | Standalone, but in a Symfony app
| Swiftmailer version | 6.2.3 (same in current master)
| PHP version         | 7.3

### Observed behaviour
From the code in `Swift_SmtpTransport` and `Swift_Transport_EsmtpTransport`, I understand that that
`Swift_SmtpTransport::__construct` allows one of `$encryption = null | "ssl" | "tls"`.

Can you please explain what the difference between "ssl" and "tls" is?
I'm only guessing, it could be:
1. "ssl" = SSLv3, "tls" = TLSv1 or later
2. "ssl" = SMTPS, "tls" = SMTP + STARTTLS

### Expected behaviour
The parameter should be documented.

### Example

```php
    /**
     * @param string $host
     * @param int    $port
     * @param string $encryption    // - put explanation here -
     */
    public function __construct($host = 'localhost', $port = 25, $encryption = null)
```

Thanks :-)

heidemn:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Doc update?   | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1329 
| License       | MIT

Fixes #1329 
Got help from #545

Both the PHPDoc comments and the documentation are unclear about the difference between `ssl` and `tls`.
Especially considering that `ssl` nowadays sounds "old & deprecated", this needed clarification, since in this case
`ssl` is surprisingly the more secure option.

### DIFF
--- a/doc/sending.rst
+++ b/doc/sending.rst
@@ -95,7 +95,8 @@ within your application and adjust the settings accordingly if the code is
 moved or if the SMTP server is changed.
 
 Some SMTP servers -- Google for example -- use encryption for security reasons.
-Swift Mailer supports using both SSL and TLS encryption settings.
+Swift Mailer supports using both ``ssl`` (SMTPS = SMTP over TLS) and ``tls``
+(SMTP with STARTTLS) encryption settings.
 
 Using the SMTP Transport
 ^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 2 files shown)


heidemn:
@fabpot please review if you have time... thanks!

fabpot:
Thank you @heidemn.

PR #1352, opened 2021-08-18
Add information about EOM

fabpot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Doc update?   | yes
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets | n/a 
| License       | MIT

I tried to work on the next major version of Swiftmailer for many years. But I never managed to finish anything meaningful as the number of changes was just too important (from adopting modern PHP features like namespaces to implement new "best practices", and remove the many special ways that makes Swiftmailer work not so well).

So, a couple of years ago, I've decided to start again, but with a different approach. Instead of modernizing the code one step at a time, I've copied the code and refactored it as much as possible. The result is Symfony Mailer.

Now that the feature set of Symfony Mailer is the same as Switfmailer, It's time to officially abandon the maintenance of Swiftmailer in favor of Symfony Mailer.

Why November 2021? Because that's the release of Symfony 5.4 (LTS) and 6.0.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+**Swiftmailer will stop being maintained at the end of November 2021.**
+
+Please, move to Symfony Mailer at your earliest convenience.
+Symfony Mailer is the next evolution of Swiftmailer.
+It provides the same features with support for modern PHP code and support for third-party providers.
+See https://symfony.com/doc/current/mailer.html for more information.
+
 6.2.8 (2021-XX-XX)
 ------------------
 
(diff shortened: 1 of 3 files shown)


nicolas-grekas (approved):

derrabus (approved):

fabpot:
More information here: https://symfony.com/blog/the-end-of-swiftmailer

Salvador1313 (commented):
Спасибо, возня со мной наверное утомительна. Думаю освоюсь со временем. Пока что смотрю на всё как баран на новые ворота.

